PR #30, opened 2018-06-19
Implement Azure storage as resource auto config and sample

warrenzhu25:


spring-cloud-azure-autoconfigure/src/main/java/com/microsoft/azure/spring/cloud/autoconfigure/storage/AzureStorageAutoConfiguration.java
```diff
@@ -35,17 +38,17 @@
 @ConditionalOnClass(CloudStorageAccount.class)
 @ConditionalOnProperty(name = "spring.cloud.azure.storage.account")
 @EnableConfigurationProperties(AzureStorageProperties.class)
+@Import(AzureStorageProtocolResolver.class)
```
yaweiw: Why Import? Is it being used here?

spring-cloud-azure-autoconfigure/src/main/java/com/microsoft/azure/spring/cloud/autoconfigure/storage/AzureStorageAutoConfiguration.java
```diff
@@ -35,17 +38,17 @@
...
 public class AzureStorageAutoConfiguration {
```
yaweiw: Shouldn't it be AzureStorageAccountAutoConfiguration since it wraps storage account object here.

spring-cloud-azure-autoconfigure/src/main/java/com/microsoft/azure/spring/cloud/autoconfigure/storage/AzureStorageAutoConfiguration.java
warrenzhu25: This import means that we need this class registered as bean for later usage. We rely on this resolver to handle azure storage specific url

spring-cloud-azure-autoconfigure/src/main/java/com/microsoft/azure/spring/cloud/autoconfigure/storage/AzureStorageAutoConfiguration.java
warrenzhu25: `Storage` in this class name is more like a general concept, rather than a specific thing. Although currently we only provide a bean of type `storageAccount`, we might provide more storage related beans in the future.

spring-cloud-azure-autoconfigure/src/main/java/com/microsoft/azure/spring/cloud/autoconfigure/storage/AzureStorageAutoConfiguration.java
warrenzhu25: According to https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/context/annotation/Import.html, this import also allows for regular component classes, which register the imported class as bean in application context. I have confirmed this, other ways such as `@Component` and configuration class provide such bean don't work neither

PR #34, opened 2018-06-27
Eventhub refactor

warrenzhu25:


spring-cloud-azure-eventhub-stream-binder/src/test/java/com/microsoft/azure/eventhub/stream/binder/EventHubTestOperation.java
```diff
@@ -25,46 +24,35 @@
...
 
     @Override
-    public Subscriber<EventData> subscribe(String eventHubName, String consumerGroup) {
-        return new TestSubscriber(eventHubName, consumerGroup);
-    }
-
-    private class TestSubscriber implements Subscriber<EventData> {
-        private final String eventHubName;
-        private final String consumerGroup;
+    public boolean subscribe(String destination, Consumer<Iterable<EventData>> consumer, String consumerGroup) {
```
yaweiw: does this method always return true? how do we handle exception here?

spring-cloud-azure-eventhub-stream-binder/src/test/java/com/microsoft/azure/eventhub/stream/binder/EventHubTestOperation.java
```diff
@@ -25,46 +24,35 @@
...
+        consumerMap.get(destination).remove(consumerGroup);
+        return true;
+    }
 
-        @Override
-        public Checkpointer<EventData> getCheckpointer() {
-            return new EventHubTestCheckpointer();
-        }
+    @Override
+    public Checkpointer<EventData> getCheckpointer(String destination, String consumerGroup) {
```
yaweiw: destination and consumerGroup are not used. why added?

warrenzhu25: Same as above for testing purpose and simplicity.

spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/core/SubscribingOperation.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+public interface SubscribingOperation<D> {
+
+    /**
+     * Register a message consumer to a given destination.
+     *
+     * @return {@code true} if the consumer was subscribed or {@code false} if it
+     * was already subscribed.
+     */
+    boolean subscribe(String destination, Consumer<Iterable<D>> consumer);
```
sophiaso: Why use `boolean` return value to indicate subscribe operation status? It requires the user to check the subscribe result with if/else conditions, why not we throw wrapped exception here instead to indicate the operation fails?

spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/core/SubscribingOperation.java
```diff
@@ -0,0 +1,38 @@
...
+
+    /**
+     * Un-register a message consumer.
+     *
+     * @return {@code true} if the consumer was un-registered, or {@code false}
```
sophiaso: Similar with above, though return value is `void`, but the Javadoc writes the return value is boolean.

spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/core/SendingOperation.java
```diff
@@ -0,0 +1,23 @@
...
+import java.util.concurrent.CompletableFuture;
+
+
+/**
+ * Operations for sending message to a destination.
+ *
+ * @param <D> the type of message
+ * @author Warren Zhu
+ */
+public interface SendingOperation<D> {
```
sophiaso: Why naming the interface name with `-ing` format?

spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/core/SubscribingByGroupOperation.java
```diff
@@ -0,0 +1,38 @@
...
+package com.microsoft.azure.spring.integration.core;
+
+import java.util.function.Consumer;
+
+/**
+ * Operations for subscribing to a destination with a consumer group.
+ *
+ * @author Warren Zhu
+ */
+public interface SubscribingByGroupOperation<D> {
```
sophiaso: Same `-ing` format as above.

spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/core/SendingOperation.java
yaweiw: Send or SendMessage?

spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/core/SubscribingByGroupOperation.java
yaweiw: same

sophiaso: According to the class name,  is there other subscription type? e.g., by non-group operation? Wonder should we define subscription in the same class instead of divide into different classes for group operation and non-group operation.

spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/eventhub/EventHubTemplate.java
```diff
@@ -0,0 +1,143 @@
...
+ * <p>
+ * The main eventhub component for sending to and consuming from event hub
+ *
+ * @author Warren Zhu
+ */
+public class EventHubTemplate implements EventHubOperation {
+
+    private static final Log LOGGER = LogFactory.getLog(EventHubTemplate.class);
+    private final ConcurrentHashMap<Tuple<String, String>, Set<Consumer<Iterable<EventData>>>>
+            consumersByNameAndConsumerGroup = new ConcurrentHashMap<>();
```
yaweiw: consumerToNameConsumerGroupMap?

spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/core/SendingOperation.java
yaweiw: Do we know type of message at compile time? Wondering if a generic interface is needed here.

spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/eventhub/EventHubTemplate.java
```diff
@@ -0,0 +1,143 @@
...
+            } else if (!Strings.isNullOrEmpty(partitionSupplier.getPartitionKey())) {
+                return client.send(message, partitionSupplier.getPartitionKey());
+            } else {
+                return client.send(message);
+            }
+        } catch (EventHubRuntimeException e) {
+            LOGGER.error(String.format("Failed to send to '%s' ", eventHubName), e);
+            CompletableFuture<Void> future = new CompletableFuture<>();
+            future.completeExceptionally(e);
+            return future;
```
sophiaso: Why the exception is not thrown out here, and instead, only logged?

spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/core/SendingOperation.java
```diff
@@ -0,0 +1,23 @@
...
+    /**
+     * Send a message to the given destination with a given partition supplier.
+     */
+    CompletableFuture<Void> sendAsync(String destination, D message, PartitionSupplier partitionSupplier);
```
yaweiw: D should be EventData.

spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/core/SubscribingByGroupOperation.java
```diff
@@ -0,0 +1,38 @@
...
+
+    /**
+     * Register a message consumer to a given destination with a given consumer group.
+     *
+     * @return {@code true} if the consumer was subscribed or {@code false} if it
+     * was already subscribed.
+     */
+    boolean subscribe(String destination, Consumer<Iterable<D>> consumer, String consumerGroup);
```
yaweiw: I'd rather make interface method simplist possible. e.g. avoid generics if can.

spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/eventhub/EventHubTemplate.java
```diff
@@ -0,0 +1,143 @@
...
+
+    @Override
+    public synchronized boolean subscribe(String destination, Consumer<Iterable<EventData>> consumer,
+            String consumerGroup) {
+        Tuple<String, String> nameAndConsumerGroup = new Tuple<>(destination, consumerGroup);
+        consumersByNameAndConsumerGroup.putIfAbsent(nameAndConsumerGroup, new CopyOnWriteArraySet<>());
+        boolean added = consumersByNameAndConsumerGroup.get(nameAndConsumerGroup).add(consumer);
+
+        if (!added) {
+            return false;
```
sophiaso: Wonder for the user who calls this subscribe method, and returned false for an `added` failure operation, the user will have no idea why the operation failed, and will not be able to recover/process such situation, also the method itself leaves no hint why the subscribe failed.

Same as above, personally prefer exception over returning boolean status.

spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/core/SubscribingOperation.java
warrenzhu25: In normal case, users doesn't need to check return result if they doesn't care the idempotent caused by subscribing same consumer multiple times. Exception has no benefits over this.

spring-cloud-azure-eventhub-stream-binder/src/test/java/com/microsoft/azure/eventhub/stream/binder/EventHubTestOperation.java
warrenzhu25: This is mocked EventHubOperation for testing purpose, so just return true.

spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/core/SubscribingOperation.java
warrenzhu25: Typo here, should be boolean

spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/eventhub/EventHubTemplate.java
```diff
@@ -0,0 +1,143 @@
...
+    @Override
+    public synchronized boolean unsubscribe(String destination, Consumer<Iterable<EventData>> consumer,
+            String consumerGroup) {
+        Tuple<String, String> nameAndConsumerGroup = new Tuple<>(destination, consumerGroup);
+        boolean existed = consumersByNameAndConsumerGroup.get(nameAndConsumerGroup).remove(consumer);
+        if (consumersByNameAndConsumerGroup.get(nameAndConsumerGroup).isEmpty()) {
+            processorHostsByNameAndConsumerGroup.remove(nameAndConsumerGroup).unregisterEventProcessor();
+        }
+
+        return existed;
```
sophiaso: Same boolean here, and `existed true` for `removed successful`?

spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/core/SubscribingByGroupOperation.java
warrenzhu25: You should have seen `SubscribingOperation` before. Combining will cause some methods throwing unsupported exception. Confusing user.

spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/core/SubscribingOperation.java
sophiaso: If the user checks the subscription result and it's false, what can the user do with this?

spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/core/SendingOperation.java
warrenzhu25: In Java's convention, async method should be suffixed with Async. Generic type is used in `EventHubOperation`, we also have similar `ServiceBusOperation` which has different message type

spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/core/SendingOperation.java
warrenzhu25: Should be generic.

spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/eventhub/EventHubTemplate.java
warrenzhu25: Please check the comment, false is only returned when this consumer is already subscribed. So user doesn't need to recover, or some programming error existed.

spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/core/SubscribingOperation.java
warrenzhu25: Nothing or the user has programming error.

spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/core/SendingOperation.java
warrenzhu25: Spring also has `SendingOperation`. In the future, we might need to user spring's rather than provide our own. So keep it consistent here.

spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/core/SubscribingByGroupOperation.java
warrenzhu25: Agree.

spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/core/SubscribingByGroupOperation.java
warrenzhu25: Generic has to be used here to improve re usability. Reasons as before.

PR #35, opened 2018-06-29
Add telemetry dependency and configuration files.

Incarnation-p-lee:

### DIFF
--- a/spring-cloud-azure-autoconfigure/pom.xml
+++ b/spring-cloud-azure-autoconfigure/pom.xml
@@ -13,6 +13,10 @@
     <artifactId>spring-cloud-azure-autoconfigure</artifactId>
     <name>Spring Cloud Azure Autoconfigure</name>
 
+    <properties>
+        <azure.applicationinsights.version>2.0.1</azure.applicationinsights.version>
+    </properties>
+
     <dependencies>
         <dependency>
             <groupId>org.springframework.boot</groupId>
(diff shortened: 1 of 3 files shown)


spring-cloud-azure-autoconfigure/src/main/resources/META-INF/project.properties
```diff
@@ -0,0 +1 @@
+project.version=@project.version@
```
warrenzhu25: Why this properties is inside META-INF?

spring-cloud-azure-autoconfigure/src/main/resources/ApplicationInsights.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<ApplicationInsights xmlns="http://schemas.microsoft.com/ApplicationInsights/2013/Settings" schemaVersion="2014-05-30">
+    <!-- The key from the portal: -->
```
warrenzhu25: I don't see any useful information in this xml file. Is this file required?

spring-cloud-azure-autoconfigure/pom.xml
```diff
@@ -73,6 +77,34 @@
...
+
     </dependencies>
 
-</project>
\ No newline at end of file
+    <build>
+        <resources>
+            <!-- Filtering Resource -->
+            <resource>
+                <directory>src/main/resources</directory>
```
warrenzhu25: This filter means not including this file in building phase, right? Then why need this file?

Incarnation-p-lee: No, it will include the file and replace the `placeholder`.

spring-cloud-azure-autoconfigure/src/main/resources/META-INF/project.properties
Incarnation-p-lee: From convention here, telemetry will take this version as property.

spring-cloud-azure-autoconfigure/src/main/resources/ApplicationInsights.xml
Incarnation-p-lee: From AI SDK will use this, and only fill the Key when release.

PR #37, opened 2018-07-02
Refactor by extracting inbound and outbound adapter logic

warrenzhu25:


spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/core/AbstractAzureMessageHandler.java
```diff
@@ -0,0 +1,165 @@
...
+    protected MessageConverter messageConverter;
+    private boolean sync = false;
+    private ListenableFutureCallback<Void> sendCallback;
+    private EvaluationContext evaluationContext;
+    private Expression sendTimeoutExpression = new ValueExpression<>(DEFAULT_SEND_TIMEOUT);
+    private Expression partitionKeyExpression;
+
+    public AbstractAzureMessageHandler(String destination, SendOperation<D> sendOperation) {
+        Assert.hasText(destination, "destination can't be null or empty");
+        Assert.notNull(sendOperation, "sendOperation can't be null");
```
Incarnation-p-lee: How about `@NonNull` from springframework ?

spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/core/Memoizer.java
```diff
@@ -0,0 +1,27 @@
...
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Function;
+
+/**
+ * Memorize function execution result
+ *
+ * @author Warren Zhu
+ */
+public class Memoizer {
```
sophiaso: `Memoizer` typo? And the method names below too.

warrenzhu25: I followed from https://opencredo.com/lambda-memoization-in-java-8/ and https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/concurrent/Memoizer.html

yaweiw: public final class Memoizer?

yaweiw: also add a private constructor to this class

spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/core/Tuple.java
```diff
@@ -12,11 +12,23 @@
     private final T first;
     private final U second;
 
-    public Tuple(T first, U second) {
+    private Tuple(T first, U second) {
```
yaweiw: interesting to see Java doesn't have Tuple implemented in SDK. Wanted to bring it up for discussion: from design perspective, do you think this is better https://msdn.microsoft.com/en-us/library/system.tuple(v=vs.110).aspx? e.g. make it static, have a create method, no getFirst() method etc.

spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/eventhub/DefaultEventHubClientFactory.java
```diff
@@ -71,62 +69,72 @@ public void setCheckpointStorageAccountContainer(String checkpointStorageAccount
...
-
-            try {
-                return EventHubClient
-                        .createSync(getOrCreateConnectionString(eventHubName), Executors.newSingleThreadExecutor());
-            } catch (EventHubException | IOException e) {
-                throw new EventHubRuntimeException("Error when creating event hub client", e);
-            }
-        });
+    public Function<String, EventHubClient> getEventHubClientCreator() {
+        return Memoizer.memoize(clientsByName, this::createEventHubClient);
```
yaweiw: can this:: be removed?

spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/eventhub/DefaultEventHubClientFactory.java
```diff
@@ -34,30 +37,25 @@
  */
 public class DefaultEventHubClientFactory implements EventHubClientFactory, DisposableBean {
     private static final Log LOGGER = LogFactory.getLog(DefaultEventHubClientFactory.class);
-    // eventHubName -> eventHubClient
-    private final ConcurrentHashMap<String, EventHubClient> clientMap = new ConcurrentHashMap<>();
 
-    // eventHubName -> connectionString
-    private final ConcurrentHashMap<String, String> connectionStringMap = new ConcurrentHashMap<>();
+    private final Map<String, EventHubClient> clientsByName = new ConcurrentHashMap<>();
```
yaweiw: it's a 1 to 1 mapping so i guess clientByName better?

spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/core/AbstractAzureMessageHandler.java
yaweiw: i think we should be consistent across all our Java code base e.g. annotation, getter setter etc.

spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/core/Memoizer.java
Incarnation-p-lee: try `@NoArgsConstructor(access = AccessLevel.PRIVATE)` here.

spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/eventhub/EventHubTemplate.java
```diff
@@ -57,12 +57,13 @@ public EventHubTemplate(EventHubClientFactory clientFactory) {
         Assert.hasText(eventHubName, "eventHubName can't be null or empty");
         Assert.notNull(message, "message can't be null");
         try {
-            EventHubClient client = this.clientFactory.getOrCreateEventHubClient(eventHubName);
+            EventHubClient client = this.clientFactory.getEventHubClientCreator().apply(eventHubName);
```
yaweiw: Wondering why we'd introduce this Memoizer and Function? I indeed think clientFactory.getOrCreateEventHubClient is better than clientFactory.getEventHubClientCreator().apply.

yaweiw: Memoization is used to cache expensive resources. EventHubClient isn't cacheable as it is tightly coupled with connectionstring and eventhub name. A new eventHubName means a new creation.
https://www.javatips.net/api/commons-lang-master/src/main/java/org/apache/commons/lang3/concurrent/Memoizer.java

spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/core/AbstractAzureMessageHandler.java
warrenzhu25: Agree.

spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/eventhub/DefaultEventHubClientFactory.java
warrenzhu25: no. It's method reference syntax.

spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/eventhub/DefaultEventHubClientFactory.java
warrenzhu25: Yes.

spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/eventhub/EventHubTemplate.java
warrenzhu25: 1. Memoizer is used to remove duplication of `putIfAbsent` operation of map. Another benefit is to separate the concern of different function by taking adavantage of functional composition.
2. EventHubClient is only coupled by eventhub name. So we can use Memoizer to cache with name as input and client as output.

spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/core/Tuple.java
warrenzhu25: I have static factory method `of`.

spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/eventhub/EventHubTemplate.java
yaweiw: Good learning to me. Thanks.

PR #39, opened 2018-07-02
Add telemetry mode for spring cloud azure autoconfiguration, #38.

Incarnation-p-lee:
* track version, installId, subsId, resourcGroup and serviceName.

* add helper class hashMac and property loader.


spring-cloud-azure-autoconfigure/src/main/java/com/microsoft/azure/spring/cloud/autoconfigure/telemetry/GetHashMac.java
```diff
@@ -0,0 +1,128 @@
...
+import java.security.NoSuchAlgorithmException;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.Locale;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public class GetHashMac {
```
warrenzhu25: 1. What's the purpose of this class? Could you add comments to show this?
2. Class name is normally a noun. `GetHashMac` is not a noun. And it's unclear why we need this class. Why mac?

Incarnation-p-lee: for obtaining the hashed value of MAC address, we use this as the unique id for each machine here. `HashMac` may be a better name her, will refine this.

spring-cloud-azure-autoconfigure/src/main/java/com/microsoft/azure/spring/cloud/autoconfigure/telemetry/PropertyLoader.java
```diff
@@ -0,0 +1,65 @@
...
+
+    private static final String PROJECT_PROPERTY_FILE = "/META-INF/project.properties";
+    private static final String APPLICATION_PROPERTY_FILE = "/application.properties";
+    private static final String APPLICATION_YML_FILE = "/application.yml";
+
+    public static String getProjectVersion() {
+        return getPropertyByName("project.version", PROJECT_PROPERTY_FILE);
+    }
+
+    public static boolean isApplicationTelemetryAllowed() {
```
warrenzhu25: Could we try to take advantage of standard `@ConfigurationProperties`, rather than implementing our own?
I think we can use this property to determine whether do auto config.

Incarnation-p-lee: yes, we can add field for telemetry in `AzureProperties`, but not for ProjectVersion.

spring-cloud-azure-autoconfigure/src/main/java/com/microsoft/azure/spring/cloud/autoconfigure/telemetry/PropertyLoader.java
```diff
@@ -0,0 +1,65 @@
...
+
+    private static String getPropertyByName(@NonNull String name, @NonNull String filename) {
+        final Properties properties = new Properties();
+        final InputStream inputStream = PropertyLoader.class.getResourceAsStream(filename);
+
+        if (inputStream == null) {
+            return null;
+        }
+
+        try {
```
warrenzhu25: Could you use `try-with-resources` instead?

spring-cloud-azure-autoconfigure/src/main/java/com/microsoft/azure/spring/cloud/autoconfigure/telemetry/PropertyLoader.java
```diff
@@ -0,0 +1,65 @@
...
+        } else {
+            return telemetryAllowed.equalsIgnoreCase("false") ? false : true;
+        }
+    }
+
+    private static String getPropertyByName(@NonNull String name, @NonNull String filename) {
+        final Properties properties = new Properties();
+        final InputStream inputStream = PropertyLoader.class.getResourceAsStream(filename);
+
+        if (inputStream == null) {
```
warrenzhu25: Returning null is confusing here, since user couldn't distinguish between property is null or resource file not existed.

spring-cloud-azure-autoconfigure/src/main/java/com/microsoft/azure/spring/cloud/autoconfigure/telemetry/PropertyLoader.java
```diff
@@ -0,0 +1,65 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Properties;
+
+@Log
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public class PropertyLoader {
+
+    private static final String PROJECT_PROPERTY_FILE = "/META-INF/project.properties";
+    private static final String APPLICATION_PROPERTY_FILE = "/application.properties";
```
warrenzhu25: I'm unsure whether introducing another property file just for access to project version is a good idea. I find this, but don't know whether it works. https://stackoverflow.com/questions/2712970/get-maven-artifact-version-at-runtime

spring-cloud-azure-autoconfigure/src/main/java/com/microsoft/azure/spring/cloud/autoconfigure/telemetry/GetHashMac.java
```diff
@@ -0,0 +1,128 @@
...
+            }
+        } catch (IOException e) {
+            return null;
+        }
+
+        return macBuilder.toString();
+    }
+
+    private static String getHexDigest(byte digest) {
+        final String hex = Integer.toString((digest & 0xff) + 0x100, 16);
```
warrenzhu25: What do these magic numbers mean?

spring-cloud-azure-autoconfigure/src/main/java/com/microsoft/azure/spring/cloud/autoconfigure/telemetry/TelemetryProperties.java
```diff
@@ -0,0 +1,23 @@
...
+ * Licensed under the MIT License. See LICENSE in the project root for
+ * license information.
+ */
+package com.microsoft.azure.spring.cloud.autoconfigure.telemetry;
+
+import lombok.AccessLevel;
+import lombok.NoArgsConstructor;
+
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public class TelemetryProperties {
```
warrenzhu25: As I see, these constants are only used by TelemetryTracker. Could we move it into that class and make them private?

spring-cloud-azure-autoconfigure/src/main/java/com/microsoft/azure/spring/cloud/autoconfigure/telemetry/TelemetryTracker.java
```diff
@@ -0,0 +1,65 @@
...
+
+import static com.microsoft.azure.spring.cloud.autoconfigure.telemetry.TelemetryProperties.*;
+
+public class TelemetryTracker {
+
+    private static final String PROJECT_VERSION = TelemetryTracker.class.getPackage().getImplementationVersion();
+
+    private static final String PROJECT_INFO = "spring-cloud-azure" + "/" + PROJECT_VERSION;
+
+    private TelemetryClient client;
```
warrenzhu25: Make it final, and initialize new client here.

spring-cloud-azure-autoconfigure/src/main/java/com/microsoft/azure/spring/cloud/autoconfigure/telemetry/TelemetryTracker.java
```diff
@@ -0,0 +1,65 @@
...
+
+    private Azure azure;
```
warrenzhu25: Make it final.

spring-cloud-azure-autoconfigure/src/main/java/com/microsoft/azure/spring/cloud/autoconfigure/telemetry/TelemetryProperties.java
Incarnation-p-lee: make sense, will move it to tracker.

spring-cloud-azure-autoconfigure/src/main/java/com/microsoft/azure/spring/cloud/autoconfigure/telemetry/TelemetryTracker.java
```diff
@@ -0,0 +1,65 @@
...
+
+    public TelemetryTracker(Azure azure) {
+        this.client = new TelemetryClient();
+        this.azure = azure;
+    }
+
+    public void trackEvent(@NonNull String name, Map<String, String> customProperties) {
```
warrenzhu25: If this is only used inside this class, could we make it private and customProperties as `@NonNull`?

spring-cloud-azure-autoconfigure/src/main/java/com/microsoft/azure/spring/cloud/autoconfigure/telemetry/TelemetryTracker.java
```diff
@@ -0,0 +1,65 @@
...
+        final Map<String, String> properties;
+        final Map<String, String> defaultProperties = this.getDefaultProperties();
+
+        if (customProperties == null) {
```
warrenzhu25: If customProperties is never null, we can merge all these into one statement.

spring-cloud-azure-autoconfigure/src/main/java/com/microsoft/azure/spring/cloud/autoconfigure/telemetry/TelemetryTracker.java
```diff
@@ -0,0 +1,65 @@
...
+    public void trackEventWithServiceName(@NonNull String eventName, @NonNull String serviceName) {
+        final Map<String, String> properties = new HashMap<>();
+
+        properties.put(PROPERTY_SERVICE_NAME, serviceName);
+
+        this.trackEvent(eventName, properties);
+    }
+
+    private Map<String, String> getDefaultProperties() {
+        final Map<String, String> properties = new HashMap<>();
```
warrenzhu25: I suppose if method is called multiple times, each time a new duplicate map will be created. Could we cache it?

spring-cloud-azure-autoconfigure/src/main/java/com/microsoft/azure/spring/cloud/autoconfigure/telemetry/TelemetryTracker.java
Incarnation-p-lee: I am not sure if we need to provide more than one track function in this class. Or just keep one public method is enough here.

spring-cloud-azure-autoconfigure/src/main/java/com/microsoft/azure/spring/cloud/autoconfigure/telemetry/TelemetryUtils.java
```diff
@@ -0,0 +1,48 @@
...
+import java.net.NetworkInterface;
+import java.net.SocketException;
+import java.net.UnknownHostException;
+
+import static com.microsoft.applicationinsights.core.dependencies.apachecommons.codec.digest.DigestUtils.sha256Hex;
+
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public class TelemetryUtils {
+
+    private static final String UNKNOWN_MAC = "unknown-Mac-Address";
```
warrenzhu25: Unknown-Mac-Address or unknown-mac-address

spring-cloud-azure-autoconfigure/src/main/java/com/microsoft/azure/spring/cloud/autoconfigure/telemetry/TelemetryUtils.java
```diff
@@ -0,0 +1,48 @@
...
+
+    private static String getMacAddress() {
+        final InetAddress ip;
+        final NetworkInterface network;
+        final byte[] macBytes;
+
+        try {
```
warrenzhu25: Could you merge declaration inside try clause, and just return in normal case?

spring-cloud-azure-autoconfigure/src/main/java/com/microsoft/azure/spring/cloud/autoconfigure/telemetry/TelemetryTracker.java
Incarnation-p-lee: telemetry only trigger one time for each module, I think it is too early to worry about performance.

spring-cloud-azure-autoconfigure/src/main/java/com/microsoft/azure/spring/cloud/autoconfigure/telemetry/TelemetryUtils.java
Incarnation-p-lee: will refine this.

spring-cloud-azure-autoconfigure/src/main/java/com/microsoft/azure/spring/cloud/autoconfigure/telemetry/TelemetryUtils.java
```diff
@@ -0,0 +1,48 @@
...
+            ip = InetAddress.getLocalHost();
+            network = NetworkInterface.getByInetAddress(ip);
+            macBytes = network.getHardwareAddress();
+        } catch (UnknownHostException | SocketException e) { // Omit
```
warrenzhu25: I suppose omit means ignore. We could rename e as ignore.

spring-cloud-azure-autoconfigure/src/main/java/com/microsoft/azure/spring/cloud/autoconfigure/telemetry/TelemetryUtils.java
```diff
@@ -0,0 +1,44 @@
...
+import static com.microsoft.applicationinsights.core.dependencies.apachecommons.codec.digest.DigestUtils.sha256Hex;
+
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public class TelemetryUtils {
+
+    private static final String UNKNOWN_MAC = "Unknown-Mac-Address";
+
+    private static String getMacAddress() {
+        try {
+            final InetAddress ip = InetAddress.getLocalHost();
```
warrenzhu25: Sorry that I should have brought this up early. 
1.  Is localhost always ip? Could it be a hostname? If this, rename ip into localhost.
2. `network` is also confusing. Maybe we can chain method call without introducing more variable and complexity.

spring-cloud-azure-autoconfigure/src/main/java/com/microsoft/azure/spring/cloud/autoconfigure/telemetry/TelemetryUtils.java
warrenzhu25: Each declaration and initialization could be one statement inside try clause. You can check latest change which fixed this.

spring-cloud-azure-autoconfigure/src/test/java/com/microsoft/azure/spring/cloud/autoconfigure/context/AzureContextAutoConfigurationTest.java
```diff
@@ -34,15 +38,34 @@ public void testAzurePropertiesConfigured() {
                     assertThat(context.getBean(AzureProperties.class).getCredentialFilePath()).isEqualTo("credential");
                     assertThat(context.getBean(AzureProperties.class).getResourceGroup()).isEqualTo("group1");
                     assertThat(context.getBean(AzureProperties.class).getRegion()).isEqualTo("westUS");
+                    assertThat(context.getBean(TelemetryTracker.class)).isNotNull();
```
warrenzhu25: This method is testing properties configured correctly. This assert seems not part of property.

spring-cloud-azure-autoconfigure/src/test/java/com/microsoft/azure/spring/cloud/autoconfigure/context/AzureContextAutoConfigurationTest.java
```diff
@@ -34,15 +38,34 @@ public void testAzurePropertiesConfigured() {
...
+    public void testAzurePropertiesTelemetryConfigured() {
+        this.contextRunner
+                .withPropertyValues("spring.cloud.azure.credentialFilePath=credential")
+                .withPropertyValues("spring.cloud.azure.resourceGroup=group1")
+                .withPropertyValues("spring.cloud.azure.region=westUS")
+                .withPropertyValues("spring.cloud.azure.telemetryAllowed=true")
+                .run(context -> assertThat(context.getBean(TelemetryTracker.class)).isNotNull());
+    }
+
+    @Test(expected = NoSuchBeanDefinitionException.class)
```
warrenzhu25: Will doesNotHaveBean work? Rather than rely on specific exception?

spring-cloud-azure-autoconfigure/src/test/java/com/microsoft/azure/spring/cloud/autoconfigure/context/AzureContextAutoConfigurationTest.java
Incarnation-p-lee: OK. I create new test for this.

PR #55, opened 2018-07-09
use slf4j logger

sophiaso:
## Description
Replace commons logging with slf4j, in order to allow users to select preferred log library.


sophiaso:
refs #50

Issue #59, opened 2018-07-09
Inject application insight instrument key as build property


warrenzhu25:
1. Check whether we can remove ApplicationInsights.xml.
2. Inject instrument key as build property.

PR #60, opened 2018-07-10
Inject ai instrumentation key as application.properties, #59.

Incarnation-p-lee:
## Description
* Remove some unused files.
* Add telemetery properties.

## Related PRs
List related PRs against other branches:

branch | PR
------ | ------
other_pr_production | [link]()
other_pr_master | [link]()

## Steps to Test
Steps to test code change


spring-cloud-azure-autoconfigure/pom.xml
```diff
@@ -15,6 +15,8 @@
 
     <properties>
         <azure.applicationinsights.version>2.0.1</azure.applicationinsights.version>
+        <!-- Instrumentation Key will be filled by maven args -Dai.instrumentationKey=xxxx for release -->
+        <ai.instrumentationKey>${place-holder}</ai.instrumentationKey>
```
warrenzhu25: Maybe this is not needed. Just inject property into application.property file.

spring-cloud-azure-autoconfigure/src/main/java/com/microsoft/azure/spring/cloud/autoconfigure/context/AzureContextAutoConfiguration.java
```diff
@@ -57,7 +58,8 @@ public Azure azure() throws IOException {
 
     @Bean
     @ConditionalOnProperty(name = "spring.cloud.azure.telemetryAllowed", havingValue = "true", matchIfMissing = true)
```
warrenzhu25: Could we add another condition for instrument key? Only if it existed do we trigger the telemetry. This is needed when other contributor want to build this project without passing instrument key

spring-cloud-azure-autoconfigure/src/main/java/com/microsoft/azure/spring/cloud/autoconfigure/telemetry/TelemetryProperties.java
```diff
@@ -0,0 +1,19 @@
...
+
+package com.microsoft.azure.spring.cloud.autoconfigure.telemetry;
+
+import lombok.Getter;
+import lombok.Setter;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+@Getter
+@Setter
+@ConfigurationProperties("ai")
```
warrenzhu25: I prefer `spring.cloud.azure.telemetry.key`. If we change this, then `spring.cloud.azure.telemetryAllowed` might be changed into `spring.cloud.azure.telemetry.enabled` to keep consistency.

spring-cloud-azure-autoconfigure/src/main/java/com/microsoft/azure/spring/cloud/autoconfigure/telemetry/TelemetryTracker.java
```diff
@@ -32,8 +33,11 @@
 
     private final Map<String, String> defaultProperties;
 
-    public TelemetryTracker(Azure azure, String resourceGroup) {
+    private TelemetryProperties telemetryProperties;
```
warrenzhu25: final

spring-cloud-azure-autoconfigure/src/main/java/com/microsoft/azure/spring/cloud/autoconfigure/telemetry/TelemetryTracker.java
```diff
@@ -43,10 +47,15 @@ public TelemetryTracker(Azure azure, String resourceGroup) {
...
     private void trackEvent(@NonNull String name, @NonNull Map<String, String> customProperties) {
-        this.defaultProperties.forEach(customProperties::putIfAbsent);
+        final String instrumentationKey = this.telemetryProperties.getInstrumentationKey();
+
+        if (StringUtils.hasText(instrumentationKey)) {
+            this.defaultProperties.forEach(customProperties::putIfAbsent);
 
-        this.client.trackEvent(name, customProperties, null);
-        this.client.flush();
+            this.client.getContext().setInstrumentationKey(instrumentationKey);
```
warrenzhu25: I think this only needed once in constructor, rather than in each event.

spring-cloud-azure-autoconfigure/pom.xml
Incarnation-p-lee: will remove this.

spring-cloud-azure-autoconfigure/src/main/java/com/microsoft/azure/spring/cloud/autoconfigure/context/AzureContextAutoConfiguration.java
Incarnation-p-lee: just for recording here, 2 `ConditionalOnProperty` is not allowed for single bean, and we may need to define new `Conditional` Annotation here. So keep current for now.

spring-cloud-azure-autoconfigure/src/main/java/com/microsoft/azure/spring/cloud/autoconfigure/telemetry/TelemetryTracker.java
```diff
@@ -42,11 +44,27 @@ public TelemetryTracker(Azure azure, String resourceGroup) {
...
+
+        if (StringUtils.hasText(instrumentationKey)) {
+            final TelemetryClient client = new TelemetryClient();
+
+            client.getContext().setInstrumentationKey(instrumentationKey);
+
+            return client;
+        }
+
+        return null;
```
warrenzhu25: Please don't return null directly, how about throw runtime exception?

spring-cloud-azure-autoconfigure/src/main/java/com/microsoft/azure/spring/cloud/autoconfigure/telemetry/TelemetryTracker.java
```diff
@@ -42,11 +44,27 @@ public TelemetryTracker(Azure azure, String resourceGroup) {
         this.defaultProperties.put(PROPERTY_INSTALLATION_ID, TelemetryUtils.getHashMac());
     }
 
+    private TelemetryClient getTelemetryClient(TelemetryProperties telemetryProperties) {
+        final String instrumentationKey = telemetryProperties.getInstrumentationKey();
+
+        if (StringUtils.hasText(instrumentationKey)) {
```
warrenzhu25: If instrument key has special format, do we need to validate here? If passed wrong key, will application insight sdk report error? Also, we should log validation error here, otherwise we're unaware of such error.

Incarnation-p-lee: I am not sure if there is a format here. insight sdk take care the invalid key silently. As the key is only available around use, I think we may not need to handle that at all.

spring-cloud-azure-autoconfigure/src/main/java/com/microsoft/azure/spring/cloud/autoconfigure/telemetry/TelemetryTracker.java
Incarnation-p-lee: I think runtime exception may not suitable here, because telemetry logic should not effect on normal code.

spring-cloud-azure-autoconfigure/src/main/java/com/microsoft/azure/spring/cloud/autoconfigure/telemetry/TelemetryTracker.java
```diff
@@ -42,11 +52,27 @@ public TelemetryTracker(Azure azure, String resourceGroup) {
...
+            client.getContext().setInstrumentationKey(TELEMETRY_INVALID_KEY);
+            LOG.warn("Telemetry instrumentationKey {} is invalid", instrumentationKey);
+        }
+
+        return client;
+    }
+
     private void trackEvent(@NonNull String name, @NonNull Map<String, String> customProperties) {
-        this.defaultProperties.forEach(customProperties::putIfAbsent);
+        if (!this.client.getContext().getInstrumentationKey().equals(TELEMETRY_INVALID_KEY)) {
```
warrenzhu25: What does this mean?

Incarnation-p-lee: If there key is invalid, we may not send the trigger. And it will be set in constructor depends on key values.

PR #62, opened 2018-07-10
Comment the not used field of PR template

Incarnation-p-lee:
## Description
Comment the not used field of PR template

## Related PRs
List related PRs against other branches:


samukce:
If it's not used, Why don't remove instead of comment it?

PR #63, opened 2018-07-10
Refactor Spring cloud azure auto-config

warrenzhu25:


spring-cloud-azure-autoconfigure/src/main/java/com/microsoft/azure/spring/cloud/autoconfigure/telemetry/TelemetryTracker.java
```diff
@@ -31,55 +32,69 @@
...
+    private TelemetryClient buildTelemetryClient(String instrumentationKey) {
+        TelemetryClient client = new TelemetryClient();
 
-        if (StringUtils.hasText(instrumentationKey) && instrumentationKey.length() == INSTRUMENTATION_KEY_LENGTH) {
-            client.getContext().setInstrumentationKey(instrumentationKey);
-        } else {
-            client.getContext().setInstrumentationKey(TELEMETRY_INVALID_KEY);
+        if (!isValid(instrumentationKey)) {
             LOG.warn("Telemetry instrumentationKey {} is invalid", instrumentationKey);
+            throw new RuntimeException("Telemetry instrumentationKey is invalid");
```
Incarnation-p-lee: From my side, I don't think it should throw runtime exception here.

1. telemetry is for our BI data, in fact, we three are almost the only users for this module.
2. Too easy for this runtime exception, if config BI allowed, and without this key in maven command line will trigger this.
3. For any other users, they do not honor this, all about telemetry to them is `telemetryAllowed`.

spring-cloud-azure-autoconfigure/src/main/java/com/microsoft/azure/spring/cloud/autoconfigure/context/AzureContextAutoConfiguration.java
```diff
@@ -59,7 +60,8 @@ public Azure azure() throws IOException {
     @Bean
     @ConditionalOnProperty(name = "spring.cloud.azure.telemetryAllowed", havingValue = "true", matchIfMissing = true)
     public TelemetryTracker telemetryTracker(Azure azure, AzureProperties azureProperties,
-                                             TelemetryProperties telemetryProperties) {
-        return new TelemetryTracker(azure, azureProperties.getResourceGroup(), telemetryProperties);
+            TelemetryProperties telemetryProperties) {
+        return new TelemetryTracker(azure.getCurrentSubscription().subscriptionId(), azureProperties.getResourceGroup(),
+                    telemetryProperties.getInstrumentationKey());
```
Incarnation-p-lee: passing properties instead of `key`, if there may be another property from telemetry, then you cannot add new args.

spring-cloud-azure-autoconfigure/src/main/java/com/microsoft/azure/spring/cloud/autoconfigure/sql/SqlServerJdbcDataSourcePropertiesUpdater.java
```diff
@@ -33,13 +33,11 @@ String getUserName() {
...
     String getUrl(DataSourceProperties dataSourceProperties) {
-        SqlServer sqlServer =
-            azureAdmin.getOrCreateSqlServer(azureSqlProperties.getServerName(), dataSourceProperties.getUsername(),
-                    dataSourceProperties.getPassword());
-        azureAdmin.createSqlDatabaseIfNotExists(azureSqlProperties.getServerName(),
+        SqlServer sqlServer = azureAdmin
+                .getOrCreateSqlServer(azureSqlProperties.getServerName(), dataSourceProperties.getUsername(),
+                        dataSourceProperties.getPassword());
+        azureAdmin.getOrCreateSqlDatabase(azureSqlProperties.getServerName(), azureSqlProperties.getDatabaseName());
+        return String.format(databaseType.getJdbcUrlTemplate(), sqlServer.fullyQualifiedDomainName(),
```
Incarnation-p-lee: could you please add some local variable to hold something like `azureSqlProperties.getServerName()` ? From my side, wrap lines is difficult to read.

spring-cloud-azure-autoconfigure/src/main/java/com/microsoft/azure/spring/cloud/autoconfigure/context/AzureContextAutoConfiguration.java
warrenzhu25: As I see for now, key is only property required for initializing this tracker. So I prefer not rely on properties.

spring-cloud-azure-autoconfigure/src/main/java/com/microsoft/azure/spring/cloud/autoconfigure/telemetry/TelemetryTracker.java
warrenzhu25: 1. Whether this class is used internally or externally, we should throw such exception since without correct key, this instance is useless. We need to throw it to let user to be aware of it.
2. Yes. So we might need catch this exception in other place. In this PR, I just merged the conflict. 
3. Yes. This key is transparent for end users. Even user build the project without this, this should run clean.

Incarnation-p-lee: Well, I am OK with exception, my point is that there must be trade-off somewhere. From my side, involve exception looks too heavy here.

PR #81, opened 2018-07-11
Use network hostname as event hub hostNamePrefix, #76.

Incarnation-p-lee:
## Description
* Get hostname from java network.

## Steps to Test
Steps to test code change

### DIFF
--- a/spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/eventhub/DefaultEventHubClientFactory.java
+++ b/spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/eventhub/DefaultEventHubClientFactory.java
@@ -102,7 +102,7 @@ public class DefaultEventHubClientFactory implements EventHubClientFactory, Disp
     }
 
     private EventProcessorHost createEventProcessorHost(Tuple<String, String> nameAndConsumerGroup) {
-        return new EventProcessorHost(EventProcessorHost.createHostName("hostNamePrefix"),
+        return new EventProcessorHost(EventProcessorHost.createHostName(HostnameHelper.getHostname()),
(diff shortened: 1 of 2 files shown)


spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/eventhub/DefaultEventHubClientFactory.java
```diff
@@ -102,7 +102,7 @@ private PartitionSender createPartitionSender(Tuple<EventHubClient, String> clie
     }
 
     private EventProcessorHost createEventProcessorHost(Tuple<String, String> nameAndConsumerGroup) {
-        return new EventProcessorHost(EventProcessorHost.createHostName("hostNamePrefix"),
```
warrenzhu25: Could we store hostname in a static field without getting it each time?

spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/eventhub/HostnameHelper.java
```diff
@@ -0,0 +1,33 @@
...
+import java.net.UnknownHostException;
+
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public class HostnameHelper {
+
+    private static final String UNKNOWN_HOST_NAME = "Unknown-HostName";
+
+    private static final String HOST_NAME = getNetworkHostname();
+
+    private static String getNetworkHostname() {
```
warrenzhu25: I think `getHostName` is good enough, `network` here is confusing.

Incarnation-p-lee: As there is public method outside take this name, I add details information for where the hostname from.

Issue #84, opened 2018-07-11
Run azure cache sample in command line cannot locate the credential file.


Incarnation-p-lee:
## Branch
* master

## Your Environment
* Version used:
* Operating System and version (desktop or mobile):
* SDK version:

PR #86, opened 2018-07-12
fix sample commandline failed to locate credential file, #84.

Incarnation-p-lee:
## Description
* fix sample commandline failed to locate credential file
* Create temp file and copy data from inputStream.

## Steps to Test
Steps to test code change

### DIFF
--- a/spring-cloud-azure-context/pom.xml
+++ b/spring-cloud-azure-context/pom.xml
@@ -23,17 +23,15 @@
             <artifactId>spring-context-support</artifactId>
             <optional>true</optional>
         </dependency>
+        <dependency>
+            <groupId>commons-io</groupId>
+            <artifactId>commons-io</artifactId>
+        </dependency>
 
         <dependency>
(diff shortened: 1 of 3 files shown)


spring-cloud-azure-context/pom.xml
```diff
@@ -28,5 +28,12 @@
             <groupId>com.microsoft.azure</groupId>
             <artifactId>azure</artifactId> <!-- We may not need include all azure management sdks in a batch -->
         </dependency>
+
+        <dependency>
+            <groupId>commons-io</groupId>
+            <artifactId>commons-io</artifactId>
+            <version>2.6</version>
+            <scope>compile</scope>
```
sophiaso: Can we move this dependency to the bom `spring-cloud-azure-dependencies` and remove the version from here?

Incarnation-p-lee: good point, will refine it.

PR #92, opened 2018-07-13
Use consistent enable property for all auto config modules

warrenzhu25:


spring-cloud-azure-autoconfigure/src/main/java/com/microsoft/azure/spring/cloud/autoconfigure/context/AzureContextAutoConfiguration.java
```diff
@@ -30,14 +26,18 @@
  * @author Warren Zhu
  */
 @Configuration
-@EnableConfigurationProperties({AzureProperties.class, TelemetryProperties.class})
-@ConditionalOnClass(name = "com.microsoft.azure.management.Azure")
-@ConditionalOnProperty("spring.cloud.azure.credentialFilePath")
+@EnableConfigurationProperties(AzureProperties.class)
+@ConditionalOnClass(Azure.class)
+@ConditionalOnProperty("spring.cloud.azure.enabled")
```
Incarnation-p-lee: context should be always enabled ?

warrenzhu25: Any auto config could be disabled. And user just provide their bean to continue use other dependent parts

Incarnation-p-lee: By default, we should enable all modules, right ?

spring-cloud-azure-autoconfigure/src/test/java/com/microsoft/azure/spring/cloud/autoconfigure/context/AzureContextAutoConfigurationTest.java
```diff
@@ -25,23 +26,27 @@
 
     @Test
     public void testAzurePropertiesConfigured() {
-        this.contextRunner.withPropertyValues("spring.cloud.azure.credentialFilePath=credential")
+        this.contextRunner.
+                                  withPropertyValues("spring.cloud.azure.credentialFilePath=credential")
```
Incarnation-p-lee: should be aligned ?

warrenzhu25: It seems auto formatted by IntelliJ

PR #100, opened 2018-07-16
 Use functional composition to log azure resource creation

warrenzhu25:


spring-cloud-azure-context/src/main/java/com/microsoft/azure/spring/cloud/context/core/AzureAdmin.java
```diff
@@ -207,27 +232,50 @@ private ServiceBusSubscription createServiceBusTopicSubscription(Tuple<Topic, St
...
             if (result != null) {
                 return result;
             }
 
-            return creator.apply(t);
+            return logCreator(creator, resourceType).apply(t);
+        };
+    }
+
+    private <T, R> Function<T, R> logCreator(Function<T, R> creator, Class<R> resourceType) {
```
Incarnation-p-lee: Could u please help to try another name rather than `logCreator`? I am not sure if I understand correctly, here we do resource creation, and add some log here. The name `logCreator` looks it will create a log.

PR #170, opened 2018-08-07
 Change Checkpoint type parameter as checkpoint key type

warrenzhu25:


spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/eventhub/DefaultEventHubClientFactory.java
```diff
@@ -50,8 +51,7 @@
     private final EventHubNamespace namespace;
     private String checkpointStorageConnectionString;
 
-    public DefaultEventHubClientFactory(AzureAdmin azureAdmin, String namespace) {
-        Assert.notNull(azureAdmin, "azureAdmin can't be null.");
+    public DefaultEventHubClientFactory(@NonNull AzureAdmin azureAdmin, String namespace) {
```
Incarnation-p-lee: suggest to use `@NonNull` from `org.springframework.lang.NonNull`.

warrenzhu25: What's the diff between lombok nonnull and spring?

Incarnation-p-lee: for stability consideration, some lombok may conflict with some maven plugin. If there are two similar annotation, I prefer the one from springframework.

PR #189, opened 2018-08-22
Add Azure message converter

warrenzhu25:


spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/core/converter/AbstractAzureMessageConverter.java
```diff
@@ -0,0 +1,101 @@
...
+import java.io.IOException;
+import java.util.Arrays;
+
+/**
+ * Abstract class handles common conversion logic between <T> and {@link Message}
+ *
+ * @author Warren Zhu
+ */
+public abstract class AbstractAzureMessageConverter<T> implements AzureMessageConverter<T> {
+    private static ObjectMapper objectMapper = new ObjectMapper();
```
sophiaso: Whether we need to consider allow user to customize this `objectMapper`, in order to support different data mapping requirements?

spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/core/converter/AbstractAzureMessageConverter.java
```diff
@@ -0,0 +1,101 @@
...
+        return (Message<U>) internalToMessage(azureMessage, headers, targetPayloadClass);
+    }
+
+    protected abstract byte[] getPayload(T azureMessage);
+
+    protected abstract T fromString(String payload);
+
+    protected abstract T fromByte(byte[] payload);
+
+    protected void setCustomHeaders(Message<?> message, T azureMessage) {
```
sophiaso: This method is empty, add `TODO`?

spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/core/converter/AbstractAzureMessageConverter.java
```diff
@@ -0,0 +1,101 @@
...
+
+    private <U> Message<?> internalToMessage(T azureMessage, MessageHeaders headers, Class<U> targetPayloadClass) {
+        byte[] payload = getPayload(azureMessage);
+
+        if (targetPayloadClass.isInstance(azureMessage)) {
+            return MessageBuilder.withPayload(azureMessage).copyHeaders(headers).build();
+        }
+
+        if (targetPayloadClass == String.class) {
+            return MessageBuilder.withPayload(new String(payload)).copyHeaders(headers).build();
```
sophiaso: Why the compare here is `exactly equals` String.class (same below), but internalFromMessage is checking with `instance of`?

spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/core/converter/AbstractAzureMessageConverter.java
```diff
@@ -0,0 +1,101 @@
...
+/**
+ * Abstract class handles common conversion logic between <T> and {@link Message}
+ *
+ * @author Warren Zhu
+ */
+public abstract class AbstractAzureMessageConverter<T> implements AzureMessageConverter<T> {
+    private static ObjectMapper objectMapper = new ObjectMapper();
+
+    @Override
+    public T fromMessage(Message<?> message, Class<T> targetClass) {
```
yaweiw: Would message conversion failure be possible? Shall we have a MessageConversionException anyway?

spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/core/converter/AbstractAzureMessageConverter.java
yaweiw: I think it is a hook method users can override.

spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/core/converter/AbstractAzureMessageConverter.java
```diff
@@ -0,0 +1,101 @@
...
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public <U> Message<U> toMessage(T azureMessage, MessageHeaders headers, Class<U> targetPayloadClass) {
+        return (Message<U>) internalToMessage(azureMessage, headers, targetPayloadClass);
+    }
+
+    protected abstract byte[] getPayload(T azureMessage);
+
+    protected abstract T fromString(String payload);
```
yaweiw: how do we handle payload that's massive? we should treat the payload in the way of stream regardless it's text message or byte message.

e.g. MarshallingMessageConverter

spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/core/converter/AbstractAzureMessageConverter.java
warrenzhu25: We could provide a setter to let users provide their `objectMapper`.  But in this version, I want to keep it simply.

spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/core/converter/AbstractAzureMessageConverter.java
warrenzhu25: Yes.

spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/core/converter/AbstractAzureMessageConverter.java
warrenzhu25: This comparison is between both Class literal. Another is between object and Class literal.

spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/core/converter/AbstractAzureMessageConverter.java
warrenzhu25: We have `ConversionException` here.

spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/core/converter/AbstractAzureMessageConverter.java
warrenzhu25: 1. If payload is too large, then backed service should throw exception. Normally, service has its constraints of message size. 
2. `Stream` also has text-based and byte-based, we still need to distinguish.

yaweiw: my ask is: we should we look into org.springframework.jms.support.converter. See if we could do better. Message conversion will be widely used in our repo.

spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/core/converter/ConversionException.java
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * @author Warren Zhu
+ */
+public class ConversionException extends NestedRuntimeException {
+
+    public ConversionException(String msg) {
+        super(msg);
+    }
+
+    public ConversionException(String msg, Throwable cause) {
```
yaweiw: cause @Nullable?

PR #190, opened 2018-08-22
Redesign subscribe operation

warrenzhu25:
1. Change subscribe to only support one consumer for each group
2. Change subscribe consumer to only support one message
3. Remove listener mode now since only record mode is supported now.


spring-cloud-azure-eventhub-stream-binder/src/main/java/com/microsoft/azure/eventhub/stream/binder/EventHubMessageChannelBinder.java
```diff
@@ -77,7 +77,6 @@ protected MessageProducer createConsumerEndpoint(ConsumerDestination destination
                 new EventHubInboundChannelAdapter(destination.getName(), this.eventHubOperation, group);
         inboundAdapter.setBeanFactory(getBeanFactory());
         // Spring cloud stream only support record mode now
```
yaweiw: remove these comments?

warrenzhu25: Yes

PR #193, opened 2018-08-27
Polish change of subscribe operation

warrenzhu25:


spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/eventhub/EventHubTemplate.java
```diff
@@ -124,18 +124,19 @@ public boolean subscribe(String destination, String consumerGroup, Consumer<Mess
...
-                                            .whenComplete((s, t) -> {
-                                                if (t != null) {
-                                                    LOGGER.warn(String.format(
-                                                            "Failed to unregister consumer '%s' with group '%s'",
-                                                            destination, consumerGroup), t);
-                                                }
-                                            });
+        EventProcessorHost processorHost = processorByNameAndGroup.remove(nameAndConsumerGroup);
+
+        processorHost.unregisterEventProcessor().whenComplete((s, t) -> {
```
yaweiw: the thread will return without waiting for the action to complete?

warrenzhu25: Yes

PR #203, opened 2018-08-31
Add service bus queue test

warrenzhu25:


spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/core/support/ServiceBusQueueTestOperation.java
```diff
@@ -0,0 +1,86 @@
...
+        handlersByQueue.put(name, handler);
+    }
+
+    @Override
+    public boolean unsubscribe(String name) {
+        handlersByQueue.removeAll(name);
+        return true;
+    }
+
+    public static <E> Optional<E> getRandom (Collection<E> e) {
```
Incarnation-p-lee: space after method name ?

warrenzhu25: Yes, should be no space.

PR #204, opened 2018-09-03
Reorganize package and add more tests

warrenzhu25:


spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/servicebus/factory/ServiceBusConnectionStringProvider.java
```diff
@@ -0,0 +1,37 @@
...
+import com.microsoft.azure.management.servicebus.AuthorizationKeys;
+import com.microsoft.azure.management.servicebus.ServiceBusNamespace;
+import com.microsoft.azure.spring.cloud.context.core.util.Memoizer;
+import com.microsoft.azure.spring.integration.servicebus.ServiceBusRuntimeException;
+import org.springframework.lang.NonNull;
+
+import java.util.function.Function;
+
+class ServiceBusConnectionStringProvider {
+    private ServiceBusNamespace serviceBusNamespace;
```
Incarnation-p-lee: final?

warrenzhu25: Yes

PR #249, opened 2018-09-14
Refactor AzureAdmin to support disable auto resource creation

warrenzhu25:


spring-cloud-azure-context/src/main/java/com/microsoft/azure/spring/cloud/context/core/impl/RedisCacheManager.java
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    @Override
+    String getResourceName(String key) {
+        return key;
+    }
+
+    @Override
+    String getResourceType() {
+        return RedisCache.class.getSimpleName();
```
Incarnation-p-lee: I am not sure if we can define enum for ResourceType instead of class simpleNames here.

warrenzhu25: For simplicity now, no need to define another enum.

PR #266, opened 2018-09-27
Implement event hub reactive operation

warrenzhu25:


spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/eventhub/AbstractEventHubTemplate.java
```diff
@@ -0,0 +1,111 @@
...
+    private CheckpointMode checkpointMode = CheckpointMode.BATCH;
+
+    AbstractEventHubTemplate(EventHubClientFactory clientFactory) {
+        this.clientFactory = clientFactory;
+    }
+
+    private static EventProcessorOptions buildEventProcessorOptions(StartPosition startPosition) {
+        EventProcessorOptions options = EventProcessorOptions.getDefaultOptions();
+
+        if (startPosition == StartPosition.EARLISET) {
```
Incarnation-p-lee: suggest to use enum with `Function` here.

spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/eventhub/AbstractEventHubTemplate.java
```diff
@@ -0,0 +1,111 @@
...
+
+    protected void register(Tuple<String, String> nameAndGroup, EventHubProcessor eventProcessor) {
+        EventProcessorHost host = this.clientFactory.getProcessorHostCreator().apply(nameAndGroup);
+        host.registerEventProcessorFactory(context -> eventProcessor, buildEventProcessorOptions(startPosition));
+    }
+
+    protected void unregister(Tuple<String, String> nameAndGroup) {
+        this.clientFactory.getProcessorHostCreator().apply(nameAndGroup).unregisterEventProcessor()
+                          .whenComplete((s, t) -> {
+                              if (t != null) {
```
Incarnation-p-lee: seems this piece of code pattern appears more than one place, wonder if we can add one `Function` for this.

spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/eventhub/AbstractEventHubTemplate.java
warrenzhu25: what do you mean by this?

spring-integration-azure/src/main/java/com/microsoft/azure/spring/integration/eventhub/AbstractEventHubTemplate.java
warrenzhu25: only log might not need another function

PR #283, opened 2018-10-11
 Change storage queue support auto-create-resources property

warrenzhu25:


spring-cloud-azure-context/src/main/java/com/microsoft/azure/spring/cloud/context/core/impl/StorageQueueManager.java
```diff
@@ -0,0 +1,83 @@
...
+    @Override
+    public CloudQueue internalGet(Tuple<String, String> key) {
+        CloudQueueClient queueClient = this.queueClientCreator.apply(key.getFirst());
+
+        try {
+            CloudQueue cloudQueue = queueClient.getQueueReference(key.getSecond());
+            if (!cloudQueue.exists()) {
+                return null;
+            }
+            return cloudQueue;
```
Incarnation-p-lee: As it is simple if-else, how about `return cloudQueue.exists() ? cloudQueue : null` ?

PR #288, opened 2018-10-17
Implement partition count based on checkpoint for event hub template

warrenzhu25:


spring-integration-azure/spring-integration-eventhub/src/main/java/com/microsoft/azure/spring/integration/eventhub/EventHubCheckpointManager.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    void onMessage(PartitionContext context, EventData eventData) {
+        if (this.checkpointConfig.getCheckpointMode() == CheckpointMode.RECORD) {
+            context.checkpoint(eventData).whenComplete(this::checkpointHandler);
+        } else if (this.checkpointConfig.getCheckpointMode() == CheckpointMode.PARTITION_COUNT) {
+            String partitionId = context.getPartitionId();
+            this.countByPartition.putIfAbsent(partitionId, new AtomicInteger(1));
+            AtomicInteger count = this.countByPartition.get(partitionId);
+            if (count.incrementAndGet() >= checkpointConfig.getCheckpointCount()) {
```
xscript: Is it possible that multiple messages hit here at the same time?
Not an expert on EventHub. So want to double confirm,

spring-integration-azure/spring-integration-eventhub/src/test/java/com/microsoft/azure/spring/integration/eventhub/EventHubOperationSendSubscribeTest.java
```diff
@@ -63,7 +64,8 @@ protected void verifyCheckpointFailureCalled(int times) {
 
     @Test
     public void testSendReceiveWithBatchCheckpointMode() {
-        sendSubscribeOperation.setCheckpointMode(CheckpointMode.BATCH);
+        sendSubscribeOperation.setCheckpointConfig(CheckpointConfig.builder().checkpointMode(CheckpointMode.BATCH).build
+                ());
```
xscript: nit: format?

spring-integration-azure/spring-integration-eventhub/src/main/java/com/microsoft/azure/spring/integration/eventhub/EventHubTemplate.java
```diff
@@ -32,7 +32,8 @@ public boolean subscribe(String destination, String consumerGroup, Consumer<Mess
             Class<?> messagePayloadType) {
         if (subscribedNameAndGroup.putIfAbsent(Tuple.of(destination, consumerGroup), true) == null) {
             this.register(destination, consumerGroup,
-                    new EventHubProcessor(consumer, messagePayloadType, getCheckpointMode(), getMessageConverter()));
+                    new EventHubProcessor(consumer, messagePayloadType, getCheckpointConfig(), getMessageConverter
+                            ()));
```
xscript: nit: format?

spring-integration-azure/spring-integration-eventhub/src/main/java/com/microsoft/azure/spring/integration/eventhub/EventHubCheckpointManager.java
warrenzhu25: For each partition, the message is serialized so the count is for each partition.

spring-integration-azure/spring-integration-eventhub/src/test/java/com/microsoft/azure/spring/integration/eventhub/EventHubOperationSendSubscribeTest.java
warrenzhu25: what do you mean by this?

xscript: brackets should be put together with the method call?

warrenzhu25: it's in the same line, the line break is caused by github view

spring-integration-azure/spring-integration-eventhub/src/main/java/com/microsoft/azure/spring/integration/eventhub/EventHubCheckpointManager.java
```diff
@@ -0,0 +1,60 @@
...
+    void completeBatch(PartitionContext context) {
+        if (this.checkpointConfig.getCheckpointMode() == CheckpointMode.BATCH) {
+            context.checkpoint().whenComplete(this::checkpointHandler);
+        }
+    }
+
+    private void checkpointHandler(Void v, Throwable t) {
+        if (t != null) {
+            log.warn("Failed to checkpoint", t);
+        }
```
xscript: nit: also add a trace log for successful checkpoint? Might be helpful for trouble shooting. But for `RECORD` mode, it might lead to too many logs.

spring-integration-azure/spring-integration-eventhub/src/test/java/com/microsoft/azure/spring/integration/eventhub/EventHubOperationSendSubscribeTest.java
Incarnation-p-lee: from my experience, put the PR review window to the biggest monitor...

spring-integration-azure/spring-integration-eventhub/src/main/java/com/microsoft/azure/spring/integration/eventhub/EventHubCheckpointManager.java
```diff
@@ -0,0 +1,60 @@
...
+    EventHubCheckpointManager(CheckpointConfig checkpointConfig) {
+        this.checkpointConfig = checkpointConfig;
+    }
+
+    void onMessage(PartitionContext context, EventData eventData) {
+        if (this.checkpointConfig.getCheckpointMode() == CheckpointMode.RECORD) {
+            context.checkpoint(eventData).whenComplete(this::checkpointHandler);
+        } else if (this.checkpointConfig.getCheckpointMode() == CheckpointMode.PARTITION_COUNT) {
+            String partitionId = context.getPartitionId();
+            this.countByPartition.putIfAbsent(partitionId, new AtomicInteger(1));
```
xscript: Does it mean we will create a new `AtomicInteger` object for every message?
Also curious whether there is a `GetOrCreate` method for `ConcurrentHashMap`?

warrenzhu25: Should use `computeIfAbsent`, then pass lazy lambda. Then only new instance when absent.

spring-integration-azure/spring-integration-eventhub/src/main/java/com/microsoft/azure/spring/integration/eventhub/EventHubCheckpointManager.java
warrenzhu25: Since the checkpoint result doesn't contain useful info such as which message. So no benefits for successful trace log.

xscript: I see.
Then is there a way for user to know that checkpoint is happening as expected? If so, we don't have to add it.

warrenzhu25: If no error message, then checkpoint is successful.

xscript: That's fine.
Personally I prefer explicit logs because when used in a custom application, it might be possible that   our packages are not set up correctly and not working at all. But it should be very rare.

PR #290, opened 2018-10-23
Enable certificate and secret authentication with Key Vault

xscript:
## Description
Enable authentication with Azure Key Vault by following approaches:
- PEM Certificate
- Client Secret


spring-cloud-azure-keyvault-config/src/main/java/com/microsoft/azure/spring/cloud/keyvault/config/auth/Credentials.java
```diff
@@ -0,0 +1,35 @@
...
+package com.microsoft.azure.spring.cloud.keyvault.config.auth;
+
+import lombok.Getter;
+import lombok.Setter;
+import org.springframework.core.io.Resource;
+import org.springframework.validation.annotation.Validated;
+
+import javax.validation.constraints.NotEmpty;
+
+/**
```
warrenzhu25: Unsure whether we should call this `Credential`. It seems only one credential here.

spring-cloud-azure-keyvault-config/src/main/java/com/microsoft/azure/spring/cloud/keyvault/config/auth/Credentials.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+@Getter
+@Setter
+@Validated
+public class Credentials {
+
+    @NotEmpty
+    private String clientId;
+
+    private String clientSecret;
```
warrenzhu25: Is secret and certificate both required or both optional?
If only one is needed, which takes high priority?

spring-cloud-azure-keyvault-config/src/main/java/com/microsoft/azure/spring/cloud/keyvault/config/auth/Credentials.java
xscript: `clientSecret` and `clientCertificate` are different credentials.

spring-cloud-azure-keyvault-config/src/main/java/com/microsoft/azure/spring/cloud/keyvault/config/auth/Credentials.java
xscript: One of them are required. `clientCertificate` will be used if it is present. Otherwise `clientSecret` will be used.
If none of them are configured, exception will be thrown.
The resolution order will be documented.

spring-cloud-azure-keyvault-config/src/main/java/com/microsoft/azure/spring/cloud/keyvault/config/auth/AuthenticationExecutor.java
```diff
@@ -0,0 +1,17 @@
...
+ */
+
+package com.microsoft.azure.spring.cloud.keyvault.config.auth;
+
+import com.microsoft.aad.adal4j.AuthenticationContext;
+import com.microsoft.aad.adal4j.AuthenticationResult;
+
+import java.util.concurrent.Future;
+
+public interface AuthenticationExecutor {
```
warrenzhu25: 1. How about use functional programming to represent actual acquire token method? Then you don't need such inheritance hierarchy.
2. Suggest to use `CompletableFuture` rather `Future`. The later only support blocking get.

spring-cloud-azure-keyvault-config/src/main/java/com/microsoft/azure/spring/cloud/keyvault/config/auth/AadKeyVaultCredentials.java
```diff
@@ -0,0 +1,48 @@
...
+    public String doAuthenticate(String authorization, String resource, String scope) {
+        AuthenticationResult result;
+
+        //Starts a service to fetch access token.
+        ExecutorService service = null;
+        try {
+            service = Executors.newSingleThreadExecutor();
+            AuthenticationContext context = new AuthenticationContext(authorization, false, service);
+            result = authExecutor.acquireToken(context, resource).get(TIMEOUT_IN_SECONDS, TimeUnit.SECONDS);
+        } catch (Exception ex) {
```
warrenzhu25: 1. How about catching specific exception here?
2. Maybe we can have a dedicated thread pool for such auth. For each method call, creating and destroying a thread pool looks too heavy.

spring-cloud-azure-keyvault-config/src/main/java/com/microsoft/azure/spring/cloud/keyvault/config/auth/AuthenticationExecutor.java
xscript: >1. How about use functional programming to represent actual acquire token method? Then you don't need such inheritance hierarchy.
Sure. Let me try it.
>2. Suggest to use CompletableFuture rather Future. The later only support blocking get.
It comes from ADAL.  We are just wrapping it for different credentials.

spring-cloud-azure-keyvault-config/src/main/java/com/microsoft/azure/spring/cloud/keyvault/config/auth/AadKeyVaultCredentials.java
xscript: > 1. How about catching specific exception here?

It is doable. Just curious any concern that we want to make it specific? We actually don't care what exception it is and will re-throw it with a clearer error message.

> 2. Maybe we can have a dedicated thread pool for such auth. For each method call, creating and destroying a thread pool looks too heavy.

Sounds good. Let's make the thread pool a private field of `AadKeyVaultCredentials`.

warrenzhu25: Normally, we only handle checked exception when we know how to deal with. For other runtime exception, just let them throw.

xscript: I see. Then in this case I think it is okay to catch all exceptions because we want to provide better error message to users. A run-time exception with deep stack trace and very specific error messages will cost user more time to investigate.
Thoughts?

spring-cloud-azure-keyvault-config/src/main/java/com/microsoft/azure/spring/cloud/keyvault/config/auth/AuthenticationExecutor.java
xscript: After offline sync, leave the implementation as it is for better test-ability.

spring-cloud-azure-keyvault-config/src/main/java/com/microsoft/azure/spring/cloud/keyvault/config/auth/AadKeyVaultCredentials.java
```diff
@@ -7,42 +7,37 @@
...
 /**
  * Subclass of {@link KeyVaultCredentials}.
  */
-@AllArgsConstructor
 public class AadKeyVaultCredentials extends KeyVaultCredentials {
     // TODO: make it configurable.
     private static final int TIMEOUT_IN_SECONDS = 15;
+    private static final String AUTH_FAILED = "Failed to authenticate with Azure Key Vault.";
+
+    private final ExecutorService service = Executors.newSingleThreadExecutor();
```
warrenzhu25: If we have many instances of this class, maybe static will be enough.

xscript: Right. Let's leave it for now and optimize it later when it is really necessary.

PR #304, opened 2018-11-08
add spring-cloud-azure-config module with basic auth client

sophiaso:
## Description
Add spring-cloud-azure-config module for Spring integration with Azure Config service, basic util class to send valid http request is included in this PR.

Test to be added in other PRs.

### DIFF
--- a/spring-cloud-azure-config/pom.xml
+++ b/spring-cloud-azure-config/pom.xml
@@ -11,8 +11,8 @@
     <modelVersion>4.0.0</modelVersion>
 
     <artifactId>spring-cloud-azure-config</artifactId>
-    <name>Spring Cloud Azure Config Service Configuration</name>
-    <description>Spring Cloud Azure Config Service Configuration</description>
+    <name>Spring Cloud Azure Config</name>
+    <description>Integration of Spring Cloud Config and Azure Config Service</description>
 
     <dependencies>
         <dependency>
--- a/spring-cloud-azure-config/pom.xml
+++ b/spring-cloud-azure-config/pom.xml
@@ -1,0 +1,37 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xmlns="http://maven.apache.org/POM/4.0.0"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <groupId>com.microsoft.azure</groupId>
+        <artifactId>spring-cloud-azure</artifactId>
+        <version>1.0.0.BUILD-SNAPSHOT</version>
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>spring-cloud-azure-config</artifactId>
+    <name>Spring Cloud Azure Config Service Configuration</name>
+    <description>Spring Cloud Azure Config Service Configuration</description>
+
+    <dependencies>
+        <dependency>
+            <groupId>commons-codec</groupId>
+            <artifactId>commons-codec</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.httpcomponents</groupId>
+            <artifactId>httpclient</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>commons-logging</groupId>
+                    <artifactId>commons-logging</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <dependency>
+            <groupId>commons-io</groupId>
+            <artifactId>commons-io</artifactId>
+        </dependency>
+    </dependencies>
+</project>
--- a/spring-cloud-azure-config/pom.xml
+++ b/spring-cloud-azure-config/pom.xml
@@ -16,6 +16,10 @@
 
     <dependencies>
         <dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-core</artifactId>
+        </dependency>
+        <dependency>
             <groupId>commons-codec</groupId>
             <artifactId>commons-codec</artifactId>
         </dependency>
--- a/spring-cloud-azure-config/src/main/java/com/microsoft/azure/ConfigHttpClient.java
+++ b/spring-cloud-azure-config/src/main/java/com/microsoft/azure/ConfigHttpClient.java
@@ -53,7 +53,7 @@ public class ConfigHttpClient {
     public static CloseableHttpResponse execute(HttpUriRequest request, String credential, String secret)
             throws IOException, URISyntaxException {
         Map<String, String> authHeaders = buildRequestHeaders(request, new Date(), credential, secret);
-        authHeaders.forEach((k, v) -> request.setHeader(k, v));
+        authHeaders.forEach(request::setHeader);
 
         try (CloseableHttpClient httpClient = HttpClients.createDefault();
                 CloseableHttpResponse response = httpClient.execute(request)) {
--- a/spring-cloud-azure-config/src/main/java/com/microsoft/azure/ConfigHttpClient.java
+++ b/spring-cloud-azure-config/src/main/java/com/microsoft/azure/ConfigHttpClient.java
@@ -5,6 +5,14 @@
  */
 package com.microsoft.azure;
 
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.StringWriter;
+import java.net.URISyntaxException;
+import java.nio.charset.Charset;
+import java.text.SimpleDateFormat;
+import java.util.*;
+
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.codec.digest.DigestUtils;
 import org.apache.commons.codec.digest.HmacAlgorithms;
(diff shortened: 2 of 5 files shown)


spring-cloud-azure-config/src/main/java/com/microsoft/azure/ConfigHttpClient.java
```diff
@@ -0,0 +1,126 @@
...
+    private static final SimpleDateFormat GMT_DATE_FORMAT = new SimpleDateFormat(DATE_FORMAT);
+
+    static {
+        GMT_DATE_FORMAT.setTimeZone(TimeZone.getTimeZone("GMT"));
+    }
+
+    public static CloseableHttpResponse execute(HttpUriRequest request, String credential, String secret)
+            throws IOException, URISyntaxException {
+        Map<String, String> authHeaders = buildRequestHeaders(request, new Date(), credential, secret);
+        authHeaders.forEach((k, v) -> request.setHeader(k, v));
```
Incarnation-p-lee: seems can be method reference ?

sophiaso: Updated.

spring-cloud-azure-config/src/main/java/com/microsoft/azure/ConfigHttpClient.java
```diff
@@ -0,0 +1,126 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See LICENSE in the project root for
+ * license information.
+ */
+package com.microsoft.azure;
```
xscript: Package name should be `com.microsoft.azure.spring.cloud.azure.config`?

spring-cloud-azure-config/src/main/java/com/microsoft/azure/ConfigHttpClient.java
```diff
@@ -0,0 +1,126 @@
...
+        // Signature
+        byte[] decodedKey = Base64.getDecoder().decode(secret);
+        String signature = encodeHmac(HMAC_SHA_256, decodedKey, toSign);
+
+        // Compose headers
+        Map<String, String> headers = new HashMap<>();
+        headers.put("x-ms-date", requestTime);
+        headers.put("x-ms-content-sha256", contentHash);
+        headers.put("Authorization", "HMAC-SHA256 Credential=" + credential +
+                ", SignedHeaders=" + signedHeaders + ", Signature=" + signature);
```
xscript: `String.format` might be better?

spring-cloud-azure-config/src/main/java/com/microsoft/azure/ConfigHttpClient.java
```diff
@@ -0,0 +1,126 @@
...
+    private static Map<String, String> buildRequestHeaders(HttpUriRequest request, Date date, String credential,
+            String secret) throws URISyntaxException, IOException {
+        String requestTime = GMT_DATE_FORMAT.format(date);
+
+        String content = "";
+        if (request instanceof HttpEntityEnclosingRequest) {
+            content = copyInputStream(((HttpEntityEnclosingRequest) request).getEntity().getContent());
+        }
+
+        byte[] digest = new DigestUtils(SHA_256).digest(content);
```
xscript: Should we use a `static` `DigestUtils`?

spring-cloud-azure-config/pom.xml
```diff
@@ -0,0 +1,37 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xmlns="http://maven.apache.org/POM/4.0.0"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <groupId>com.microsoft.azure</groupId>
+        <artifactId>spring-cloud-azure</artifactId>
```
xscript: Should be `spring-cloud-azure-config`?

spring-cloud-azure-config/pom.xml
```diff
@@ -0,0 +1,37 @@
...
+        <version>1.0.0.BUILD-SNAPSHOT</version>
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>spring-cloud-azure-config</artifactId>
+    <name>Spring Cloud Azure Config Service Configuration</name>
```
xscript: `Spring Cloud Azure Config` should be enough. See Consul's name at [here](https://github.com/spring-cloud/spring-cloud-consul/blob/master/spring-cloud-consul-config/pom.xml#L9).
We can explain more in `description` tag though.

spring-cloud-azure-config/src/main/java/com/microsoft/azure/ConfigHttpClient.java
sophiaso: Updated to `com.microsoft.azure.config`

spring-cloud-azure-config/src/main/java/com/microsoft/azure/ConfigHttpClient.java
sophiaso: Yes, `String.format` is better.

spring-cloud-azure-config/src/main/java/com/microsoft/azure/ConfigHttpClient.java
sophiaso: Followed the [DigestUtils Javadoc](https://commons.apache.org/proper/commons-codec/apidocs/org/apache/commons/codec/digest/DigestUtils.html) to create, using static methods will also require to new some other instances like `MessageDigest`.

spring-cloud-azure-config/src/main/java/com/microsoft/azure/config/ConfigHttpClient.java
```diff
@@ -0,0 +1,128 @@
...
+    static {
+        GMT_DATE_FORMAT.setTimeZone(TimeZone.getTimeZone("GMT"));
+    }
+
+    public static CloseableHttpResponse execute(HttpUriRequest request, String credential, String secret)
+            throws IOException, URISyntaxException {
+        Map<String, String> authHeaders = buildRequestHeaders(request, new Date(), credential, secret);
+        authHeaders.forEach(request::setHeader);
+
+        try (CloseableHttpClient httpClient = HttpClients.createDefault()) {
```
warrenzhu25: If http client creation is somewhat heavy, we should cache it in static field, rather than creating and destroying for each request.

spring-cloud-azure-config/src/main/java/com/microsoft/azure/config/ConfigHttpClient.java
```diff
@@ -0,0 +1,128 @@
...
+            return httpClient.execute(request);
```
warrenzhu25: I suppose this is a blocking call to wait for response, should we make it async?

spring-cloud-azure-config/src/main/java/com/microsoft/azure/config/ConfigHttpClient.java
```diff
@@ -0,0 +1,128 @@
...
+        }
+    }
+
+    private static Map<String, String> buildRequestHeaders(HttpUriRequest request, Date date, String credential,
+            String secret) throws URISyntaxException, IOException {
+        String requestTime = GMT_DATE_FORMAT.format(date);
+
+        String content = "";
```
warrenzhu25: Suggest to extract another method only for build signature. This will make this method easier to understand.

spring-cloud-azure-config/src/main/java/com/microsoft/azure/config/ConfigHttpClient.java
```diff
@@ -0,0 +1,128 @@
...
+        String host = getHost(request);
+        String toSign = String.format("%s\n%s\n%s;%s;%s", methodName, requestPath, requestTime, host, contentHash);
+
+        // Signature
+        byte[] decodedKey = Base64.getDecoder().decode(secret);
+        String signature = encodeHmac(HMAC_SHA_256, decodedKey, toSign);
+
+        // Compose headers
+        Map<String, String> headers = new HashMap<>();
+        headers.put("x-ms-date", requestTime);
```
warrenzhu25: For some constant, suggest to put it into constant field.

spring-cloud-azure-config/src/main/java/com/microsoft/azure/config/ConfigHttpClient.java
sophiaso: Sure, will extract to another method.

spring-cloud-azure-config/src/main/java/com/microsoft/azure/config/ConfigHttpClient.java
sophiaso: Moved `CloseableHttpClient` as a instance variable.

spring-cloud-azure-config/src/main/java/com/microsoft/azure/config/ConfigHttpClient.java
sophiaso: Keep the sync version for now.

spring-cloud-azure-config/src/main/java/com/microsoft/azure/config/ConfigHttpClient.java
sophiaso: As those constants are not frequently reused, will not put into extra constant field.

PR #305, opened 2018-11-09
add config store operation interface

sophiaso:
## Description
Configuration store operation interface, which currently only includes operations might be required by configuring PropertySource.

Test is not included in this PR.


spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/ConfigServiceOperations.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+public interface ConfigServiceOperations {
+    /**
+     * Find a config key-value item with given {@code keyName} and {@link Nullable} {@code label}
+     *
+     * @param keyName name of the key
+     * @param label   label of the key, can be null
+     * @return {@link Optional<KeyValueItem>}, if key does not exist, return {@link Optional#empty()}.
+     */
+    Optional<KeyValueItem> getKey(String keyName, @Nullable String label);
```
xscript: Do we really need it?

sophiaso: Probably no requirement currently, will remove.

PR #309, opened 2018-11-13
azure config store for Spring Property Source

sophiaso:
## Description

- Azure Config Store as remote Spring Property Source
- Sample code to use the Azure Config store

### DIFF
--- a/spring-cloud-azure-config/pom.xml
+++ b/spring-cloud-azure-config/pom.xml
@@ -17,10 +17,6 @@
     <dependencies>
         <dependency>
             <groupId>org.springframework.boot</groupId>
-            <artifactId>spring-boot-autoconfigure</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-autoconfigure-processor</artifactId>
         </dependency>
         <dependency>
--- a/spring-cloud-azure-config/pom.xml
+++ b/spring-cloud-azure-config/pom.xml
@@ -16,8 +16,16 @@
 
     <dependencies>
         <dependency>
-            <groupId>org.springframework</groupId>
-            <artifactId>spring-core</artifactId>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-autoconfigure</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-autoconfigure-processor</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.cloud</groupId>
+            <artifactId>spring-cloud-context</artifactId>
         </dependency>
         <dependency>
             <groupId>commons-codec</groupId>
@@ -34,6 +42,10 @@
             </exclusions>
         </dependency>
         <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>jcl-over-slf4j</artifactId>
+        </dependency>
+        <dependency>
             <groupId>commons-io</groupId>
             <artifactId>commons-io</artifactId>
         </dependency>
@@ -44,6 +56,10 @@
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.microsoft.azure</groupId>
+            <artifactId>spring-cloud-azure-autoconfigure</artifactId>
         </dependency>
     </dependencies>
 </project>
--- a/spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/AzureCloudConfigProperties.java
+++ b/spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/AzureCloudConfigProperties.java
@@ -14,6 +14,7 @@ import org.springframework.validation.annotation.Validated;
 
 import javax.annotation.PostConstruct;
 import javax.validation.constraints.NotEmpty;
+import javax.validation.constraints.Pattern;
 
 @Validated
 @Getter
@@ -42,9 +43,10 @@ public class AzureCloudConfigProperties {
     @Nullable
     private String label;
 
-    // Path separator for the key name, e.g., /foo-app/dev/db.connection.key
+    // Profile separator for the key name, e.g., /foo-app_dev/db.connection.key
     @NotEmpty
-    private String separator = "/";
+    @Pattern(regexp = "^[a-zA-Z0-9_@]+$")
+    private String profileSeparator = "_";
 
     // Values extracted from connection string
     private String endpoint;
--- a/spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/AzureCloudConfigProperties.java
+++ b/spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/AzureCloudConfigProperties.java
@@ -10,6 +10,7 @@ import lombok.Setter;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
+import org.springframework.util.StringUtils;
 import org.springframework.validation.annotation.Validated;
 
 import javax.annotation.PostConstruct;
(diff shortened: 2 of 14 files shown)


spring-cloud-azure-samples/spring-cloud-azure-config-sample/pom.xml
```diff
@@ -0,0 +1,24 @@
...
+        <version>1.0.0.BUILD-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>spring-cloud-azure-config-sample</artifactId>
+    <name>Spring Cloud Azure Config Sample</name>
+    <dependencies>
+        <dependency>
+            <groupId>com.microsoft.azure</groupId>
+            <artifactId>spring-cloud-azure-config</artifactId>
```
xscript: We will have a `spring-cloud-starter-azure-config` package, right?

spring-cloud-azure-config/pom.xml
```diff
@@ -45,5 +57,9 @@
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
         </dependency>
+        <dependency>
```
warrenzhu25: This should already include `spring-boot-autoconfigure`

spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/AzureConfigPropertySource.java
```diff
@@ -0,0 +1,44 @@
...
+import com.microsoft.azure.spring.cloud.config.domain.KeyValueItem;
+import org.springframework.core.env.EnumerablePropertySource;
+
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+public class AzureConfigPropertySource extends EnumerablePropertySource<ConfigServiceOperations> {
+    private String context;
```
warrenzhu25: `final` ?

spring-cloud-azure-config/pom.xml
sophiaso: As it's `optional` in `spring-boot-autoconfigure`, so I think we have to explicitly include it here, though in this repo's case, these jackson library has also been transitively dependent by other azure modules.

spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/AzureConfigPropertySourceLocator.java
```diff
@@ -0,0 +1,92 @@
...
+        }
+
+        List<String> profiles = Arrays.asList(env.getActiveProfiles());
+
+        /* Generate which contexts(key prefixes) will be used for key-value items search
+           If key prefix is empty, default context is: application, current application name is: foo,
+           active profile is: dev, separator is: /
+           Will generate these contexts: /application, /application/dev, /foo, /foo/dev
+         */
+        generateContexts(this.properties.getDefaultContext(), profiles);
```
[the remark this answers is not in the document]
sophiaso: As it's only private accessed, so should have much problem, but will update.

spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/AzureConfigPropertySourceLocator.java
```diff
@@ -0,0 +1,92 @@
...
+        if (!StringUtils.hasText(applicationName)) {
+            applicationName = env.getProperty(SPRING_APP_NAME_PROP);
+        }
+
+        List<String> profiles = Arrays.asList(env.getActiveProfiles());
+
+        /* Generate which contexts(key prefixes) will be used for key-value items search
+           If key prefix is empty, default context is: application, current application name is: foo,
+           active profile is: dev, separator is: /
+           Will generate these contexts: /application, /application/dev, /foo, /foo/dev
```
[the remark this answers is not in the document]
sophiaso: Good question, need more discussion on setting rules for this case.

spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/AzureCloudConfigProperties.java
```diff
@@ -0,0 +1,57 @@
...
+
+    public String getCredential() {
+        return splitConnectionStr()[1].replaceFirst("Id=", "");
+    }
+
+    public String getSecret() {
+        return splitConnectionStr()[2].replaceFirst("Secret=", "");
+    }
+
+    private String[] splitConnectionStr() {
```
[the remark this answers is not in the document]
sophiaso: Updated to initialize only once.

spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/AzureConfigPropertySource.java
sophiaso: Updated

spring-cloud-azure-config/pom.xml
sophiaso: Removed extra dependency.

spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/AzureCloudConfigProperties.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    public String getSecret() {
+        return secret;
+    }
+
+    @PostConstruct
+    private void validateAndInit() {
+        String[] items = connectionString.split(";");
+        Assert.isTrue(items.length == 3, "Connection string should contain three items split by ;.");
```
xscript: We shouldn't care about how many items are in the connection string. All we care is the connection string contains `endpoint`, `id` and `secret`, right? Extra key vault pairs can be safely ignored. And our package will still work if service team decides to add more fields in the connection string.

spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/AzureCloudConfigProperties.java
```diff
@@ -0,0 +1,77 @@
...
+
+        this.endpoint = items[0].replaceFirst("Endpoint=", "");
```
xscript: We can not presume the order of the properties.

spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/AzureCloudConfigProperties.java
```diff
@@ -0,0 +1,93 @@
...
+
+    @PostConstruct
+    private void validateAndInit() {
+        String[] items = connectionString.split(";");
+        for (String item : items) {
+            if (!StringUtils.hasText(item)) {
+                continue;
+            }
+
+            if (item.startsWith(ENDPOINT_PREFIX)) {
```
xscript: Is `startsWith` case sensitive?

spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/AzureCloudConfigProperties.java
```diff
@@ -0,0 +1,93 @@
...
+    private String label;
+
+    // Profile separator for the key name, e.g., /foo-app_dev/db.connection.key
+    @NotEmpty
+    @Pattern(regexp = "^[a-zA-Z0-9_@]+$")
+    private String profileSeparator = "_";
+
+    // Values extracted from connection string
+    private String endpoint;
+    private String credential;
```
xscript: Why don't we name it `id` ? Align with `.NET`?

spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/AzureCloudConfigProperties.java
```diff
@@ -0,0 +1,93 @@
...
+            if (item.startsWith(ENDPOINT_PREFIX)) {
+                this.endpoint = item.substring(ENDPOINT_PREFIX.length());
+            } else if (item.startsWith(ID_PREFIX)) {
+                this.credential = item.substring(ID_PREFIX.length());
+            } else if (item.startsWith(SECRET_PREFIX)) {
+                this.secret = item.substring(SECRET_PREFIX.length());
+            }
+        }
+
+        Assert.hasText(this.endpoint, "Endpoint should not be null or empty.");
```
xscript: The error message is not clear enough. How about `Endpoint property should not be null or empty in the connection string of Azure Config Service.` ?

PR #313, opened 2018-11-14
Azure config starter

sophiaso:
## Description
Add Spring Cloud Azure Config Starter

### DIFF
--- a/spring-cloud-azure-samples/spring-cloud-azure-config-sample/pom.xml
+++ b/spring-cloud-azure-samples/spring-cloud-azure-config-sample/pom.xml
@@ -16,9 +16,5 @@
             <groupId>com.microsoft.azure</groupId>
             <artifactId>spring-cloud-starter-azure-config</artifactId>
         </dependency>
-        <dependency>
-            <groupId>org.springframework.boot</groupId>
-            <artifactId>spring-boot-starter</artifactId>
-        </dependency>
     </dependencies>
 </project>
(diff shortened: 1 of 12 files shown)


spring-cloud-azure-samples/spring-cloud-azure-config-sample/pom.xml
```diff
@@ -14,7 +14,7 @@
     <dependencies>
         <dependency>
             <groupId>com.microsoft.azure</groupId>
-            <artifactId>spring-cloud-azure-config</artifactId>
+            <artifactId>spring-cloud-starter-azure-config</artifactId>
```
warrenzhu25: With this, we should not need `spring-boot-starter`

spring-cloud-azure-samples/spring-cloud-azure-config-sample/src/main/resources/bootstrap.properties
```diff
@@ -1,4 +1,3 @@
 spring.cloud.azure.config.connection-string=put-your-connection-string-here
```
warrenzhu25: I just want to confirm whether we really need to rely on connections string. As all other spring cloud modules don't have such low level dependency, they only rely on logical concepts. And our design goal is to minimize such low level details.

spring-cloud-azure-samples/spring-cloud-azure-config-sample/pom.xml
sophiaso: Updated.

spring-cloud-azure-samples/spring-cloud-azure-config-sample/src/main/resources/bootstrap.properties
sophiaso: Will improve if possible in the future.

PR #318, opened 2018-11-15
Fix config bug

sophiaso:
## Description
Fix config bugs:

1. No restriction for key prefix
2. Null label was search as `*`, by default search null label, and `%00` for null label search, if user needs to search all labels, need to configure `spring.cloud.azure.config.label` as `*`


sophiaso:
> We will add some test cases for these functionality?

Yes, will add test cases.

PR #329, opened 2018-11-20
allow configurable order for property source locator

sophiaso:
## Description
Support user to configure the order of AzureConfigPropertySourceLocator , for how to use, refer to the test case configuration.


spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/AzureConfigPropertySourceLocator.java
```diff
@@ -99,4 +101,13 @@ private AzureConfigPropertySource create(String context) {
 
         return propertySource;
     }
+
+    @Override
+    public int getOrder() {
+        return order;
+    }
+
+    public void setOrder(int order) {
```
xscript: Is this method just for testing?

sophiaso: It's not, if user needs to configure different order, they have to use this method and create a different bean, same as the test configuration.

sophiaso: Updated to configure with properties.

PR #332, opened 2018-11-21
Fix spring boot logging level not working

warrenzhu25:


warrenzhu25:
Resend in another PR.

PR #335, opened 2018-11-22
configure project version to user agent

sophiaso:


spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/utils/PropertyLoader.java
```diff
@@ -0,0 +1,29 @@
...
+    private static final String PROJECT_PROPERTY_FILE = "/project.properties";
+
+    public static String getProjectVersion() {
+        String version = "unknown";
+        try (InputStream inputStream = PropertyLoader.class.getResourceAsStream(PROJECT_PROPERTY_FILE)) {
+            if (inputStream != null) {
+                final Properties properties = new Properties();
+                properties.load(inputStream);
+
+                version = properties.getProperty("project.version");
```
Incarnation-p-lee: not sure if `hasText` check is need here but just corner case.

pom.xml
```diff
@@ -218,6 +218,12 @@
                 </executions>
             </plugin>
         </plugins>
+        <resources>
```
warrenzhu25: Maybe you should follow other modules about how to get a project version.

sophiaso: The side effect of `Class#getPackage().getImplementationVersion()` is that, the runtime implementation will always be null.  Or we should consider it as not a problem?

sophiaso: Updated to use implementation version directly, ignore not packaged phase null version value case.

PR #340, opened 2018-11-26
Eventhub log enhancement

warrenzhu25:


spring-cloud-azure-eventhub-stream-binder/src/main/java/com/microsoft/azure/eventhub/stream/binder/properties/EventHubConsumerProperties.java
```diff
@@ -19,4 +19,5 @@
 public class EventHubConsumerProperties {
     private StartPosition startPosition = StartPosition.LATEST;
     private CheckpointMode checkpointMode = CheckpointMode.BATCH;
+    private int checkpointCount = 10;
```
Incarnation-p-lee: As it is magic number, could you please add some comments about why `10` ?

spring-integration-azure/spring-integration-azure-core/src/main/java/com/microsoft/azure/spring/integration/core/DefaultMessageHandler.java
```diff
@@ -86,12 +86,16 @@ protected void handleMessageInternal(Message<?> message) throws Exception {
     private void handleSendResponseAsync(Message<?> message, CompletableFuture<?> future) {
         future.handle((t, ex) -> {
             if (ex != null) {
-                log.warn("{} sent failed in async mode", message);
+                if (log.isWarnEnabled()) {
```
Incarnation-p-lee: just curiosity, the `log` cannot handle this logic like `if enabled` ?

spring-cloud-azure-eventhub-stream-binder/src/main/java/com/microsoft/azure/eventhub/stream/binder/properties/EventHubConsumerProperties.java
warrenzhu25: Yes, will add comment for this.

spring-integration-azure/spring-integration-azure-core/src/main/java/com/microsoft/azure/spring/integration/core/DefaultMessageHandler.java
warrenzhu25: Without if condition, even if log is disabled, log string still need to be built.

Incarnation-p-lee: I see, you are avoiding the creation of new String instance. My suggestion that may not be one issue, anyway, it is up to your choice.

warrenzhu25: Actually, it's a really big issue. Since this log will be happening for each message, this will fill memory quickly.

Incarnation-p-lee: Compiler has unique String optimization, and you may need to provide profile data about how extra memory quickly.

warrenzhu25: Each message will be different. So this optimization won't help.

PR #346, opened 2018-11-29
Support azure storage file type as resource

warrenzhu25:


spring-cloud-azure-storage/src/test/java/com/microsoft/azure/spring/cloud/storage/AzureBlobStorageTests.java
```diff
@@ -80,36 +80,36 @@ public void testWritableOutputStream() throws Exception {
 
     @Test(expected = FileNotFoundException.class)
     public void testWritableOutputStreamNoAutoCreateOnNullBlob() throws Exception {
-        String location = "blob://container/non-existing";
+        String location = "azure-blob://container/non-existing";
```
xscript: nit: could be static constant member of this class.

warrenzhu25: Yes.

PR #349, opened 2018-11-30
Resolve feedback of Eventhub binder

warrenzhu25:


spring-cloud-azure-eventhub-stream-binder/src/main/java/com/microsoft/azure/eventhub/stream/binder/config/EventHubBinderConfiguration.java
```diff
@@ -31,8 +31,6 @@
  */
 @Configuration
 @ConditionalOnMissingBean(Binder.class)
-@AutoConfigureBefore(TelemetryAutoConfiguration.class)
-@AutoConfigureAfter({AzureContextAutoConfiguration.class, AzureEventHubAutoConfiguration.class})
 @EnableConfigurationProperties({AzureEventHubProperties.class, EventHubExtendedBindingProperties.class})
```
xscript: I think we should remove the usage of `@AutoConfigureBefore` and `@AutoConfigureAfter` from the whole code base. We don't want to get the same feedback for other modules.
I am fine with that we don't fix them in this PR since this PR is all about Event Hub binder. Created #350 to track this work.

spring-cloud-azure-eventhub-stream-binder/src/main/java/com/microsoft/azure/eventhub/stream/binder/properties/EventHubConsumerProperties.java
```diff
@@ -14,10 +14,52 @@
 /**
  * @author Warren Zhu
  */
-@Getter
-@Setter
 public class EventHubConsumerProperties {
+    /**
+     * Whether the consumer receives messages from the beginning or end of event hub.
+     * If {@link StartPosition#EARLISET}, from beginning. If {@link StartPosition#LATEST}, from end.
```
xscript: `EARLISET` to `EARLIEST`?

spring-cloud-azure-eventhub-stream-binder/src/main/java/com/microsoft/azure/eventhub/stream/binder/properties/EventHubConsumerProperties.java
```diff
@@ -14,10 +14,52 @@
 /**
  * @author Warren Zhu
  */
-@Getter
-@Setter
```
xscript: Remove unnecessary dependency?

spring-cloud-azure-eventhub-stream-binder/src/main/java/com/microsoft/azure/eventhub/stream/binder/properties/EventHubExtendedBindingProperties.java
```diff
@@ -35,4 +31,13 @@ public EventHubConsumerProperties getExtendedConsumerProperties(String channelNa
     public EventHubProducerProperties getExtendedProducerProperties(String channelName) {
         return this.bindings.computeIfAbsent(channelName, key -> new EventHubBindingProperties()).getProducer();
     }
+
```
xscript: Let's remove the extra blank line.

spring-cloud-azure-eventhub-stream-binder/src/main/java/com/microsoft/azure/eventhub/stream/binder/config/EventHubBinderConfiguration.java
warrenzhu25: Agree. I will confirm all things work after removing these in another PR

spring-cloud-azure-eventhub-stream-binder/src/main/java/com/microsoft/azure/eventhub/stream/binder/properties/EventHubConsumerProperties.java
warrenzhu25: Good catch. Existed there long time.

spring-cloud-azure-eventhub-stream-binder/src/main/java/com/microsoft/azure/eventhub/stream/binder/properties/EventHubConsumerProperties.java
warrenzhu25: Yes

spring-cloud-azure-eventhub-stream-binder/src/main/java/com/microsoft/azure/eventhub/stream/binder/properties/EventHubExtendedBindingProperties.java
warrenzhu25: Yes

PR #355, opened 2018-12-03
Support retrieve connection string from ARM

sophiaso:
## Description
When connection string not configured, will try to retrieve connection string from ARM.

### DIFF
--- a/codestyle/checkstyle/checkstyle.xml
+++ b/codestyle/checkstyle/checkstyle.xml
@@ -165,7 +165,7 @@ page at http://checkstyle.sourceforge.net/config.html -->
 
             <property name="ignorePattern"
                       value="${com.puppycrawl.tools.checkstyle.checks.sizes.LineLength.ignorePattern}"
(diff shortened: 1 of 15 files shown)


codestyle/checkstyle/checkstyle.xml
```diff
@@ -165,7 +165,7 @@ page at http://checkstyle.sourceforge.net/config.html -->
 
             <property name="ignorePattern"
                       value="${com.puppycrawl.tools.checkstyle.checks.sizes.LineLength.ignorePattern}"
-                      default="^(package .*;\s*)|(import .*;\s*)|( *\* *https?://.*)$"/>
+                      default="^(package .*;\s*)|(import .*;\s*)|( *\* *https?://.*)|( *.*https?://.*)$"/>
```
warrenzhu25: What's the purpose of this change? And you might need to doc this in commit message.

sophiaso: Want to ignore comments starting with `*` and contains `http(s)://` link, in order to show in one line.

PR #360, opened 2018-12-04
support prop name is splitted by slash

sophiaso:
## Description
Slash in prop name will be converted to dots. e.g., `a/b/c` -> `a.b.c`


spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/AzureConfigPropertySource.java
```diff
@@ -42,7 +42,7 @@ public void initProperties() {
         List<KeyValueItem> items = source.getKeys(context + "*", label);
 
         for (KeyValueItem item : items) {
-            String key = item.getKey().trim().substring(context.length());
+            String key = item.getKey().trim().substring(context.length()).replace('/', '.');
```
Incarnation-p-lee: not sure previous code ensure `item` hasText.

sophiaso: Might be better to check, but it's response from remote service, basically should not happen, will consider and improve in other PR,

PR #361, opened 2018-12-04
Remove unnecessary description in homepage readme

warrenzhu25:


README.md
```diff
@@ -9,30 +9,30 @@
...
+Module | Version | Sample
+------ |--- | ---
+[Spring Cloud Stream Binder for Azure Event Hub](spring-cloud-azure-eventhub-stream-binder/) | [![Maven Central](https://img.shields.io/maven-central/v/com.microsoft.azure/spring-cloud-azure-eventhub-stream-binder.svg)](https://search.maven.org/#search%7Cga%7C1%7Ca%3A%22spring-cloud-azure-eventhub-stream-binder%22) |[Sample](https://github.com/Microsoft/spring-cloud-azure/tree/release/1.0.0.M2/spring-cloud-azure-samples/spring-cloud-azure-eventhub-binder-sample)
+[Spring Cloud Stream with Azure Event Hub Kafka API](spring-cloud-azure-samples/spring-cloud-stream-eventhub-kafka-sample/) | |[Sample](https://github.com/Microsoft/spring-cloud-azure/tree/release/1.0.0.M2/spring-cloud-azure-samples/spring-cloud-azure-eventhub-kafka-sample)
```
xscript: Suggest we also put the maven central badge here. I get the point that this is the same package as the above one. But I suspect user will get confused whether this is available.

warrenzhu25: Yes. good point.

PR #365, opened 2018-12-05
Make storage account support accessKey or credential

warrenzhu25:


spring-cloud-azure-context/src/main/java/com/microsoft/azure/spring/cloud/context/core/impl/StorageConnectionStringBuilder.java
```diff
@@ -0,0 +1,36 @@
...
+public class StorageConnectionStringBuilder {
+    private static final String DEFAULT_PROTOCOL = "DefaultEndpointsProtocol";
+
+    private static final String ACCOUNT_NAME = "AccountName";
+
+    private static final String ACCOUNT_KEY = "AccountKey";
+
+    private static final String ENDPOINT_SUFFIX = "EndpointSuffix";
+
+    private static final String HTTP_PROTOCOL = "http";
```
xscript: Is default protocol `HTTPS`?

warrenzhu25: I followed https://docs.microsoft.com/en-us/azure/storage/queues/storage-java-how-to-use-queue-storage#setup-an-azure-storage-connection-string. This link used `http`. I'll handle this issue in another PR.

xscript: Sure. You can take a look at [here](https://docs.microsoft.com/en-us/azure/storage/common/storage-security-guide#encryption-in-transit).

PR #369, opened 2018-12-06
Remove lombok

warrenzhu25:


lennartblom:
Hi there! 
Could you give statement on why `lombok` was removed from the project? 🤔 I'm just curious because of some debates in the office. Isn't the removal of boilerplate code for `getter`, `setter` and `builder` quite a win? Or are there any concerns regarding the build process or security?

Would be awesome to hear from you guys! 

Thanks in advance!

brunoborges:
We try to minimize as much as possible the number of external dependencies. If you still want to use Lombok on your own code, it is better for you to manage that dependency (which version you want), than rely on an inherited dependency.

jmax01:
Lombok is usually tagged as  `<scope>provided</scope>` and thus not a runtime dependency.
Was there a reason this wasn't done?

When marked as provided there is little impact on dependency management or security (outside of Eclipse users that need to keep their running Lombok agent version in sync with the version on the classpath).

The productivity, readability, reliability, and API ergonomics gained using Lombok, in my experience, far outweighs dealing with versions.

> We try to minimize as much as possible the number of external dependencies. If you still want to use Lombok on your own code, it is better for you to manage that dependency (which version you want), than rely on an inherited dependency.

lennartblom:
Alright. Thanks for the info guys! 🚀👍🏻🙂

brunoborges:
@jmax01 it is not about removing the dependency from the dependency graph (even with scope `provided`). It is about removing the dependency on the 3rd party library at all.

spring-cloud-azure-autoconfigure/src/main/java/com/microsoft/azure/spring/cloud/autoconfigure/telemetry/MacAddressHelper.java
```diff
@@ -6,13 +6,10 @@
 package com.microsoft.azure.spring.cloud.autoconfigure.telemetry;
 
 import com.microsoft.applicationinsights.core.dependencies.apachecommons.codec.digest.DigestUtils;
-import lombok.AccessLevel;
-import lombok.NoArgsConstructor;
 
 import java.net.InetAddress;
 import java.net.NetworkInterface;
 
-@NoArgsConstructor(access = AccessLevel.PRIVATE)
```
viktarkolbik: Why didn't you add the private constructor manually?

PR #376, opened 2018-12-11
support watch and refresh properties

sophiaso:
## Description
Add AzureCloudConfigWatch to load properties from Config service periodically.

### DIFF
--- a/spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/AzureCloudConfigAutoConfiguration.java
+++ b/spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/AzureCloudConfigAutoConfiguration.java
@@ -1,0 +1,38 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See LICENSE in the project root for
+ * license information.
+ */
+package com.microsoft.azure.spring.cloud.config;
+
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.cloud.endpoint.RefreshEndpoint;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.scheduling.TaskScheduler;
+import org.springframework.scheduling.concurrent.ThreadPoolTaskScheduler;
+
+@Configuration
+public class AzureCloudConfigAutoConfiguration {
+    public static final String WATCH_TASK_SCHEDULER_NAME = "azureConfigWatchTaskScheduler";
+
+    @Bean
+    @ConditionalOnClass(RefreshEndpoint.class)
+    @ConditionalOnProperty(name = AzureCloudConfigProperties.CONFIG_PREFIX + ".watch.enabled", matchIfMissing = true)
+    public AzureCloudConfigWatch getConfigWatch(ConfigServiceOperations operations,
+                                                AzureCloudConfigProperties properties,
+                                                @Qualifier(WATCH_TASK_SCHEDULER_NAME) TaskScheduler scheduler) {
+        return new AzureCloudConfigWatch(operations, properties, scheduler);
+    }
+
+    @Bean(name = WATCH_TASK_SCHEDULER_NAME)
+    @ConditionalOnMissingBean
+    @ConditionalOnClass(RefreshEndpoint.class)
+    @ConditionalOnProperty(name = AzureCloudConfigProperties.CONFIG_PREFIX + ".watch.enabled", matchIfMissing = true)
+    public TaskScheduler getTaskScheduler() {
+        return new ThreadPoolTaskScheduler();
+    }
+}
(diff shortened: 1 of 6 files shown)


spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/AzureCloudConfigAutoConfiguration.java
```diff
@@ -0,0 +1,38 @@
...
+import org.springframework.context.annotation.Configuration;
+import org.springframework.scheduling.TaskScheduler;
+import org.springframework.scheduling.concurrent.ThreadPoolTaskScheduler;
+
+@Configuration
+public class AzureCloudConfigAutoConfiguration {
+    public static final String WATCH_TASK_SCHEDULER_NAME = "azureConfigWatchTaskScheduler";
+
+    @Bean
+    @ConditionalOnClass(RefreshEndpoint.class)
```
warrenzhu25: Suggest to move shared condition on class level

sophiaso: Updated.

spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/AzureCloudConfigWatch.java
```diff
@@ -0,0 +1,136 @@
...
+import java.util.List;
+import java.util.Optional;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.stream.Collectors;
+
+public class AzureCloudConfigWatch implements ApplicationEventPublisherAware, SmartLifecycle {
+    private static final Logger LOGGER = LoggerFactory.getLogger(AzureCloudConfigWatch.class);
+    private final ConfigServiceOperations configOperations;
+    private final LinkedHashMap<String, String> keyNameEtagMap = new LinkedHashMap<>();
```
warrenzhu25: Why use `LinkedHashMap`? Why not use concurrent one?

sophiaso: Only one watch will be started, no concurrent is required?

warrenzhu25: If check config hasn't finished in 1 second, and then your next check has already started, then it might be on different thread.

spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/AzureCloudConfigWatch.java
```diff
@@ -0,0 +1,136 @@
...
+    public int getPhase() {
+        return 0;
+    }
+
+    public void watchConfigKeyValues() {
+        if (this.running.get()) {
+            String prefix = StringUtils.hasText(properties.getPrefix()) ? properties.getPrefix() + "*" : "*";
+            List<KeyValueItem> keyValueItems = configOperations.getKeys(prefix, properties.getLabel());
+
+            if (keyValueItems == null || keyValueItems.isEmpty()) {
```
warrenzhu25: Will null a valid return value? If not, we should not handle it.

spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/AzureCloudConfigWatch.java
```diff
@@ -0,0 +1,136 @@
...
+                LOGGER.trace("Some key matching {} is updated, will send refresh event.", prefix);
+                keyNameEtagMap.clear();
+                keyNameEtagMap.putAll(newKeyEtagMap);
+                RefreshEventData eventData = new RefreshEventData(prefix);
+                publisher.publishEvent(new RefreshEvent(this, eventData, eventData.getMessage()));
+            }
+        }
+    }
+
+    private boolean mapNotInclude(LinkedHashMap<String, String> map, String key, String value) {
```
warrenzhu25: 1. Method should be positive, rather than negative.
2. Parameter type `Map` is enough

spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/AzureCloudConfigWatch.java
```diff
@@ -0,0 +1,136 @@
...
+        return !map.containsKey(key) || (map.get(key) != null && !map.get(key).equals(value));
+    }
+
+    /**
+     * For each refresh, multiple etags can change, but even one etag is changed, refresh is required.
+     */
+    class RefreshEventData {
+        private static final String MSG_TEMPLATE = "Some key matching %s has been updated since last check.";
```
warrenzhu25: 1. Should be `some keys`
2. My suggestion is including which keys are actually changed to help troubleshooting.

sophiaso: In the [RefreshEventListener](https://github.com/spring-cloud/spring-cloud-commons/blob/master/spring-cloud-context/src/main/java/org/springframework/cloud/endpoint/event/RefreshEventListener.java#L51), the changed keys will be printed, as in here, I will publish an event without checking all keys change status, so the changed keys information is not available here.

spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/AzureCloudConfigWatch.java
sophiaso: Reasonable, updated.

spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/AzureCloudConfigWatch.java
sophiaso: Will not return null, removed null check.

spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/AzureCloudConfigWatch.java
sophiaso: Updated.

PR #379, opened 2018-12-12
query empty label as null label in config service

sophiaso:
## Description
Previously, empty label string is considered as wildcard match.

### DIFF
--- a/spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/RestAPIBuilder.java
+++ b/spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/RestAPIBuilder.java
@@ -62,7 +62,7 @@ public class RestAPIBuilder {
             this.addParam("key", prefix);
         }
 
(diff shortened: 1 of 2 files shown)


spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/RestAPIBuilder.java
```diff
@@ -62,7 +62,7 @@ public String buildKVApi(@Nullable String prefix, @Nullable String label) {
             this.addParam("key", prefix);
         }
 
-        label = label == null ? "%00" : label; // label=%00 matches null label
+        label = StringUtils.isEmpty(label) ? "%00" : label; // label=%00 matches null label
```
warrenzhu25: In another PR, you used `hasText`. Any specific reason to use `isEmpty` here

sophiaso: Should be some typo... updated.

PR #382, opened 2018-12-13
Upgrade spring cloud and boot version into 2.1.0.RC

warrenzhu25:


spring-cloud-azure-eventhub-stream-binder/src/main/java/com/microsoft/azure/eventhub/stream/binder/EventHubMessageChannelBinder.java
```diff
@@ -92,6 +92,16 @@ public EventHubProducerProperties getExtendedProducerProperties(String destinati
         return this.bindingProperties.getExtendedProducerProperties(destination);
     }
 
+    @Override
+    public String getDefaultsPrefix() {
```
sophiaso: Why there is an `s` in `Defaults`?

spring-cloud-azure-eventhub-stream-binder/src/main/java/com/microsoft/azure/eventhub/stream/binder/properties/EventHubExtendedBindingProperties.java
```diff
@@ -18,7 +19,7 @@
 @ConfigurationProperties("spring.cloud.stream.eventhub")
 public class EventHubExtendedBindingProperties
         implements ExtendedBindingProperties<EventHubConsumerProperties, EventHubProducerProperties> {
-
+    private static final String DEFAULTS_PREFIX = "spring.cloud.stream.eventhub.default";
```
sophiaso: Also an `S` in the `DEFAULTS`?

spring-cloud-azure-eventhub-stream-binder/src/main/java/com/microsoft/azure/eventhub/stream/binder/EventHubMessageChannelBinder.java
warrenzhu25: This `defaults` is define by spring cloud stream super class. As I understand, the meaning of plural is that there're many default property values.

PR #392, opened 2018-12-14
Fix cloud watch auto configuration and add test case (fix #390)

### LINKED ISSUE #390: Azconfig: spring.cloud.azure.config should not be loaded if it's disabled
I have this in bootstrap.perperties and the ConnectionString is not set.
```
spring.cloud.azure.config.enabled=false
spring.cloud.azure.config.connection-string=${ConnectionString}
```
I got error below when running in a CommandLineRunner app
```
[WARNING]
java.lang.reflect.InvocationTargetException
    at sun.reflect.NativeMethodAccessorImpl.invoke0 (Native Method)
    at sun.reflect.NativeMethodAccessorImpl.invoke (NativeMethodAccessorImpl.java:62)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke (DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke (Method.java:498)
    at org.springframework.boot.maven.AbstractRunMojo$LaunchRunner.run (AbstractRunMojo.java:558)
    at java.lang.Thread.run (Thread.java:748)
Caused by: org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'getConfigWatch' defined in class path resource [com/microsoft/azure/spring/cloud/config/AzureCloudConfigAutoConfiguration.class]: Unsatisfied dependency expressed through method 'getConfigWatch' parameter 0; nested exception is org.springframework.beans.factory.NoSuchBeanDefinitionException: No qualifying bean of type 'com.microsoft.azure.spring.cloud.config.ConfigServiceOperations' available: expected at least 1 bean which qualifies as autowire candidate. Dependency annotations: {}
```

Two issues here
1. The error message is very confusing
2. Since I have `spring.cloud.azure.config.enabled=false`, we shouldn't try to load spring.cloud.azure.config in the first place.

sophiaso:
## Description
1. Fix config not enabled, config watch being created
2. Test the cloud watch related auto configuration

refs #390

### DIFF
--- a/spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/AzureCloudConfigAutoConfiguration.java
+++ b/spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/AzureCloudConfigAutoConfiguration.java
@@ -16,22 +16,26 @@ import org.springframework.scheduling.TaskScheduler;
 import org.springframework.scheduling.concurrent.ThreadPoolTaskScheduler;
 
 @Configuration
-@ConditionalOnClass(RefreshEndpoint.class)
-@ConditionalOnProperty(prefix = AzureCloudConfigProperties.CONFIG_PREFIX, name = {"enabled", "watch.enabled"},
-        matchIfMissing = true)
+@ConditionalOnProperty(prefix = AzureCloudConfigProperties.CONFIG_PREFIX, name = "enabled", matchIfMissing = true)
 public class AzureCloudConfigAutoConfiguration {
     public static final String WATCH_TASK_SCHEDULER_NAME = "azureConfigWatchTaskScheduler";
 
-    @Bean
-    public AzureCloudConfigWatch getConfigWatch(ConfigServiceOperations operations,
-                                                AzureCloudConfigProperties properties,
-                                                @Qualifier(WATCH_TASK_SCHEDULER_NAME) TaskScheduler scheduler) {
-        return new AzureCloudConfigWatch(operations, properties, scheduler);
-    }
+    @Configuration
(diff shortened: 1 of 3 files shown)


spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/AzureCloudConfigAutoConfiguration.java
```diff
@@ -17,7 +17,8 @@
 
 @Configuration
 @ConditionalOnClass(RefreshEndpoint.class)
-@ConditionalOnProperty(name = AzureCloudConfigProperties.CONFIG_PREFIX + ".watch.enabled", matchIfMissing = true)
+@ConditionalOnProperty(prefix = AzureCloudConfigProperties.CONFIG_PREFIX, name = {"enabled", "watch.enabled"},
```
warrenzhu25: I suppose you still need to trigger this auto config when watch is disabled.

sophiaso: Currently, it only contains auto configuration for watch related beans, so when watch is disabled, no bean will be created.

warrenzhu25: Got it. Then class name might be better to contain `Watch` if this is the sole purpose.

sophiaso: OK, moved the watch related configuration into separate class, make the configuration more clear.

PR #393, opened 2018-12-14
Add service bus namespace in telemetry when only connection string provided

warrenzhu25:


spring-cloud-azure-autoconfigure/src/main/java/com/microsoft/azure/spring/cloud/autoconfigure/servicebus/ServiceBusUtils.java
```diff
@@ -0,0 +1,15 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See LICENSE in the project root for
+ * license information.
+ */
+
+package com.microsoft.azure.spring.cloud.autoconfigure.servicebus;
+
+public class ServiceBusUtils {
+    public static String getNamespace(String connectionString){
```
sophiaso: Do we need to check whether `connectionString` here is null or not?

warrenzhu25: Might handle connectionString check in event hub properties validation.

PR #398, opened 2018-12-18
keys will be searched from next page if next link header exists

sophiaso:
## Description
Support query keys from config store if next link header exists.

### DIFF
--- a/spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/ConfigServiceTemplate.java
+++ b/spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/ConfigServiceTemplate.java
@@ -48,24 +48,36 @@ public class ConfigServiceTemplate implements ConfigServiceOperations {
     public List<KeyValueItem> getKeys(@Nullable String prefix, @Nullable String label) {
         String requestUri = new RestAPIBuilder().withEndpoint(storeEndpoint).buildKVApi(prefix, label);
         List<KeyValueItem> result = new ArrayList<>();
-        CloseableHttpResponse response = getRawResponse(requestUri);
 
-        while (response != null) {
-            try {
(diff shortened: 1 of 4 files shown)


spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/ConfigServiceTemplate.java
```diff
@@ -61,4 +92,15 @@
             throw new IllegalStateException(LOAD_FAILURE_MSG, e);
         }
     }
+
+    private static String getNextLink(CloseableHttpResponse response) {
+        // Extract next link path and query from response header
```
Incarnation-p-lee: not sure java doc is better if you want to comments the method.

spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/RestAPIBuilder.java
```diff
@@ -80,7 +88,9 @@ private String buildRequestUri() {
 
         if (params != null && params.size() > 0) {
```
Incarnation-p-lee: better to use `params.isEmpty()` ?

spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/ConfigServiceTemplate.java
sophiaso: Updated Javadoc.

spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/RestAPIBuilder.java
sophiaso: Updated.

PR #400, opened 2018-12-18
Support Open Service Broker Eventhub auto config

warrenzhu25:


spring-cloud-azure-autoconfigure/src/main/java/com/microsoft/azure/spring/cloud/autoconfigure/eventhub/AzureEventHubAutoConfiguration.java
```diff
@@ -52,16 +59,38 @@ public EventHubOperation eventHubOperation(EventHubClientFactory clientFactory)
 
     @Bean
     @ConditionalOnMissingBean
-    public EventHubClientFactory clientFactory(ResourceManagerProvider resourceManagerProvider,
+    @ConditionalOnBean(ResourceManagerProvider.class)
```
sophiaso: Just curious why the `ResourceManagerProvider` bean is checked by `ConditionalOnBean`, but in the method the bean `resourceManagerProvider` is still checked? Can we put the `ResourceManagerProvider` as a not required bean?  Same for the `clientFactory` method.

warrenzhu25: 1. `ConditionalOnBean` should be removed.
2. Using optional bean is better.

PR #402, opened 2018-12-19
support multiple stores in the configuration properties

sophiaso:
## Description
Support multiple config stores in the properties, e.g.,
```yaml
spring:
  cloud:
    azure:
      config:
        stores:
          - name: my-store
            prefix: /config
            connection-string: config-store-connection-string
            label: v1
```
For the test cases and Auto Configurations, in this PR, only one config store is considered, which will not be the actual case, will improve in other PRs.

### DIFF
--- a/spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/AzureCloudConfigProperties.java
+++ b/spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/AzureCloudConfigProperties.java
@@ -63,10 +63,7 @@ public class AzureCloudConfigProperties {
             return;
         }
 
-        stores.forEach(store -> storeMap.put(store.getName(),
-                ConnectionString.of(store.getConnectionString())));
-
-        Assert.notEmpty(stores, "No config store has been configured under spring.cloud.config.stores");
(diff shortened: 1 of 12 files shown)


spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/AzureCloudConfigProperties.java
```diff
@@ -65,48 +54,19 @@
...
-                this.secret = item.substring(SECRET_PREFIX.length());
-            }
-        }
+        stores.forEach(store -> storeMap.put(store.getName(),
+                ConnectionString.of(store.getConnectionString())));
 
-        Assert.hasText(this.endpoint, String.format(NON_EMPTY_MSG, "Endpoint"));
-        Assert.hasText(this.id, String.format(NON_EMPTY_MSG, "Id"));
-        Assert.hasText(this.secret, String.format(NON_EMPTY_MSG, "Secret"));
+        Assert.notEmpty(stores, "No config store has been configured under spring.cloud.config.stores");
```
Incarnation-p-lee: seems unnecessary here, as empty will return.

PR #404, opened 2018-12-21
store connection string in separate class

sophiaso:
## Description
For easier access and more straight-forward abstraction.

### DIFF
--- a/spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/resource/ConnectionString.java
+++ b/spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/resource/ConnectionString.java
@@ -18,11 +18,11 @@ public class ConnectionString {
     public static final String NON_EMPTY_MSG = "%s property should not be null or empty in the connection string of " +
             "Azure Config Service.";
 
-    private String endpoint;
-    private String id;
-    private String secret;
+    private final String endpoint;
+    private final String id;
+    private final String secret;
 
-    public ConnectionString(String endpoint, String id, String secret) {
+    private ConnectionString(String endpoint, String id, String secret) {
         this.endpoint = endpoint;
         this.id = id;
         this.secret = secret;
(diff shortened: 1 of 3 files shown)


spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/resource/ConnectionString.java
```diff
@@ -0,0 +1,61 @@
...
+
+public class ConnectionString {
+    private static final String CONN_STRING_REGEXP = "Endpoint=https://([^;]+);Id=([^;]+);Secret=([^;]+)";
+    public static final String ENDPOINT_ERR_MSG = String.format("Connection string does not follow format %s.",
+            CONN_STRING_REGEXP);
+    private static final Pattern CONN_STRING_PATTERN = Pattern.compile(CONN_STRING_REGEXP);
+    public static final String NON_EMPTY_MSG = "%s property should not be null or empty in the connection string of " +
+            "Azure Config Service.";
+
+    private String endpoint;
```
warrenzhu25: If no change, better be `final`

spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/resource/ConnectionString.java
```diff
@@ -0,0 +1,61 @@
...
+    private String id;
+    private String secret;
+
+    public ConnectionString(String endpoint, String id, String secret) {
```
warrenzhu25: If no direct call of this constructor, better be `private`.

spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/resource/ConnectionString.java
sophiaso: updated.

spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/resource/ConnectionString.java
sophiaso: Make sense, for now there is no external access yet, use private first. Updated.

PR #416, opened 2018-12-25
support query by multiple labels and result sort by labels

sophiaso:
## Description
- Allows user configure multiple labels.
- The searched result is ordered by configured labels.

Sample configuration:
```yaml
spring:
  cloud:
    azure:
      config:
        stores:
          - name: my-store
            connection-string: my-connection-string
            label: label1, label2 
```

### DIFF
--- a/spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/RestAPIBuilder.java
+++ b/spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/RestAPIBuilder.java
@@ -67,7 +67,7 @@ public class RestAPIBuilder {
         }
 
         String label = NULL_LABEL;
-        if (labels != null) {
+        if (labels != null && !labels.isEmpty()) {
             labels = labels.stream().filter(l -> StringUtils.hasText(l))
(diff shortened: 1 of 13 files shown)


spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/RestAPIBuilder.java
```diff
@@ -53,18 +57,25 @@ public RestAPIBuilder addParam(String key, String value) {
...
-    public String buildKVApi(@Nullable String prefix, @Nullable String label) {
+    public String buildKVApi(@Nullable String prefix, @Nullable List<String> labels) {
         this.withPath(KEY_VALUE_API);
         if (StringUtils.hasText(prefix)) {
-            this.addParam("key", prefix);
+            this.addParam(KEY_PARAM, prefix);
+        }
+
+        String label = NULL_LABEL;
+        if (labels != null) {
```
Incarnation-p-lee: && !lables.isEmpty() ?

sophiaso: OK, updated.

PR #417, opened 2018-12-26
Check illegal label

sophiaso:
## Description
- Check illegal label
- Init store name if only connection string configured
- Improve validation

### DIFF
--- a/spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/AzureCloudConfigProperties.java
+++ b/spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/AzureCloudConfigProperties.java
@@ -6,16 +6,21 @@
 package com.microsoft.azure.spring.cloud.config;
 
 import com.microsoft.azure.spring.cloud.config.msi.AzureCloudConfigMSIProperties;
+import com.microsoft.azure.spring.cloud.config.resource.ConnectionString;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
+import org.springframework.util.StringUtils;
 import org.springframework.validation.annotation.Validated;
 
 import javax.annotation.PostConstruct;
 import javax.validation.constraints.NotEmpty;
 import javax.validation.constraints.Pattern;
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.stream.Collectors;
 
 @Validated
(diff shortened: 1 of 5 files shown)


spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/AzureCloudConfigProperties.java
```diff
@@ -128,18 +128,24 @@ public void setMsi(AzureCloudConfigMSIProperties msi) {
 
     @PostConstruct
     public void validateAndInit() {
+        Assert.notEmpty(this.stores, "At least one config store has to be configured.");
+
         if (!msiEnabled) {
-            Assert.notEmpty(this.stores, "At least one config store has to be configured.");
+            int uniqueStoreSize = this.stores.stream().map(s -> s.getConnectionString()).distinct()
```
xscript: Should we check config store name instead of connection string?  Connection string for the same config store may have different format?

spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/AzureCloudConfigProperties.java
```diff
@@ -128,18 +128,24 @@ public void setMsi(AzureCloudConfigMSIProperties msi) {
...
+            Assert.isTrue(this.stores.size() == uniqueStoreSize,
+                    "Duplicate connection string exists.");
+
             this.stores.forEach(store -> {
                 Assert.isTrue(StringUtils.hasText(store.getConnectionString()),
                         "Connection string cannot be empty.");
                 store.validateAndInit();
             });
+        } else {
+            int uniqueStoreSize = this.stores.stream().map(s -> s.getName()).distinct()
```
xscript: This is the right behavior IMO. Do we need a separate method to avoid duplicate code?

spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/AzureCloudConfigProperties.java
sophiaso: Check config store name only make sense, but still we do not support multi connection string for the same config store. Updated to check config store name only.

spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/AzureCloudConfigProperties.java
sophiaso: Duplicate code removed.

PR #454, opened 2019-01-16
change default polling interval to 30 seconds

sophiaso:
1. Increase default polling interval to 30 seconds
2. Change polling interval to Duration type

### DIFF
--- a/spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/AzureCloudConfigProperties.java
+++ b/spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/AzureCloudConfigProperties.java
@@ -19,6 +19,7 @@ import javax.validation.constraints.NotEmpty;
 import javax.validation.constraints.Pattern;
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.stream.Collectors;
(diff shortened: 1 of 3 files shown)


spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/AzureCloudConfigProperties.java
```diff
@@ -146,11 +147,11 @@ public void setEnabled(boolean enabled) {
             this.enabled = enabled;
         }
 
-        public int getDelay() {
+        public Duration getDelay() {
```
zhenlan: Just to confirm, the framework will convert 10s or 10ms or 10m to the right `Duration` type (10 seconds, 10 milliseconds, 10 minutes respectively) automatically and correctly?

sophiaso: Yes, it will. [Spring boot doc](https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-external-config.html#boot-features-external-config-conversion-duration)

```
ns for nanoseconds
us for microseconds
ms for milliseconds
s for seconds
m for minutes
h for hours
d for days
```

zhenlan: Awesome! Thanks.

PR #473, opened 2019-02-12
Fix RuntimeException for the watched key not exist case

sophiaso:
## Description
When watched key does not exist and watch feature enabled, the revision request with Range header will return 416 Range not Applicable error.


spring-cloud-azure-config/src/main/java/com/microsoft/azure/spring/cloud/config/ConfigServiceTemplate.java
```diff
@@ -126,6 +124,16 @@ public ConfigServiceTemplate(ConfigHttpClient httpClient, ConnectionStringPool c
         return result;
     }
 
+    private boolean itemsReachedRange(List<KeyValueItem> items, Range range) {
+        if (range == null) {
+            return false;
+        }
+
+        int maxSize = range.getEndItem() - range.getStartItem() + 1;
```
Incarnation-p-lee: is there any possibility that `endItem() < startItem()` ?

sophiaso: No, Range is not allowed to be created with `startItem <= endItem`.

PR #478, opened 2019-02-14
Rename azure config module names and properties

sophiaso:
To include appconfiguration into the package name and property names


spring-cloud-azure-samples/azure-appconfiguration-sample/src/main/java/com/example/HelloController.java
```diff
@@ -0,0 +1,18 @@
+package com.example;
```
xscript: Minor: license header?

sophiaso: Missed, will add.

spring-cloud-azure-samples/azure-appconfiguration-sample/src/main/resources/bootstrap.yaml
```diff
@@ -3,6 +3,6 @@
 spring:
   cloud:
     azure:
-      config:
-        stores:
-          - connection-string: [configuration-store-string]
+      appconfiguration:
+          stores:
```
Incarnation-p-lee: need 4 space indent ?

sophiaso: Will update the indent to align better.

PR #494, opened 2019-03-07
Remove ai sdk dependency.

Incarnation-p-lee:
* Send the ai Data from rest API.

### DIFF
--- a/spring-cloud-azure-autoconfigure/src/main/java/com/microsoft/azure/spring/cloud/autoconfigure/telemetry/TelemetrySender.java
+++ b/spring-cloud-azure-autoconfigure/src/main/java/com/microsoft/azure/spring/cloud/autoconfigure/telemetry/TelemetrySender.java
@@ -5,43 +5,56 @@
  */
 package com.microsoft.azure.spring.cloud.autoconfigure.telemetry;
 
-import com.microsoft.applicationinsights.TelemetryClient;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.http.*;
 import org.springframework.lang.NonNull;
 import org.springframework.util.StringUtils;
+import org.springframework.web.client.RestTemplate;
 
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
+
+import static org.springframework.util.MimeTypeUtils.APPLICATION_JSON;
 
 public class TelemetrySender {
     private static final Logger log = LoggerFactory.getLogger(TelemetrySender.class);
 
     private static final int INSTRUMENTATION_KEY_LENGTH = 36;
 
-    private final TelemetryClient client;
+    private static final String TELEMETRY_TARGET_URL = "https://dc.services.visualstudio.com/v2/track";
+
+    private static final ObjectMapper MAPPER = new ObjectMapper();
+
+    private static final HttpHeaders HEADERS = new HttpHeaders();
+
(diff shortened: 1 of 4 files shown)


spring-cloud-azure-autoconfigure/src/main/java/com/microsoft/azure/spring/cloud/autoconfigure/telemetry/TelemetrySender.java
```diff
@@ -5,43 +5,56 @@
...
 
+import static org.springframework.util.MimeTypeUtils.APPLICATION_JSON;
+
 public class TelemetrySender {
     private static final Logger log = LoggerFactory.getLogger(TelemetrySender.class);
 
     private static final int INSTRUMENTATION_KEY_LENGTH = 36;
 
-    private final TelemetryClient client;
+    private static final String TELEMETRY_TARGET_URL = "https://dc.services.visualstudio.com/v2/track";
```
warrenzhu25: Could you move all rest client logic into a class such as TelemetryRestClient, then just replace original client into this one? In the future, we just need to replace this client.

spring-cloud-azure-autoconfigure/src/main/java/com/microsoft/azure/spring/cloud/autoconfigure/telemetry/TelemetrySender.java
```diff
@@ -51,8 +64,36 @@ private static boolean isValid(String instrumentationKey) {
...
+
+        for (int i = 0; i < RETRY_LIMIT; i++) {
+            response = executeRequest(eventData);
+
+            if (response != null && response.getStatusCode() == HttpStatus.OK) {
+                return;
+            }
+        }
+
+        if (response != null && response.getStatusCode() != HttpStatus.OK) {
```
warrenzhu25: This if condition is not needed here.

spring-cloud-azure-autoconfigure/src/main/java/com/microsoft/azure/spring/cloud/autoconfigure/telemetry/TelemetrySender.java
```diff
@@ -51,8 +64,36 @@ private static boolean isValid(String instrumentationKey) {
...
+    private static ResponseEntity<String> executeRequest(final TelemetryEventData eventData) {
+        try {
+            final HttpEntity<String> body = new HttpEntity<>(MAPPER.writeValueAsString(eventData), HEADERS);
+
+            return REST_TEMPLATE.exchange(TELEMETRY_TARGET_URL, HttpMethod.POST, body, String.class);
+        } catch (Exception ignore) {
+            log.warn("Failed to exchange telemetry request, {}.", ignore.getMessage());
+        }
+
+        return null;
```
warrenzhu25: Suggest not returning null here.

spring-cloud-azure-autoconfigure/src/main/java/com/microsoft/azure/spring/cloud/autoconfigure/telemetry/TelemetrySender.java
Incarnation-p-lee: OK. Will refine to true or false.

spring-cloud-azure-autoconfigure/src/main/java/com/microsoft/azure/spring/cloud/autoconfigure/telemetry/TelemetrySender.java
Incarnation-p-lee: I use the same name with TelemetrySender for that purpose in other repo, but you have this name already. Will try new name for this.

PR #498, opened 2019-03-07
Add getter for telemetryEventData.

Incarnation-p-lee:
* make objectmapper happy.


Incarnation-p-lee:
The objectMapper will delete some fields if below condition fail.
```java
    public boolean anyVisible() {
        return _anyVisible(_fields)
            || _anyVisible(_getters)
            || _anyVisible(_setters)
            || _anyVisible(_ctorParameters)
        ;
    }
```

PR #529, opened 2019-07-06
Add property to enable/disable secureTransfer for azure storage

### LINKED ISSUE #528: HTTPS for Spring Cloud Azure Event Hub Stream Binder 1.0.0.M2
## Expected Behavior
* The Spring Cloud Azure EventHub Stream Binder should make HTTP**S** requests against the Storage Account related APIs. It would be desirable by default, but it's important to note that this behaviour should be configurable.

## Current Behavior
* The Spring Cloud Azure EventHub Stream Binder release 1.0.0.M2 seems to only be making HTTP requests against the Storage Account related APIs. I can't find any property that would allow me to make HTTP**S** requests instead.

The following error is obtained:
`com.microsoft.azure.storage.StorageException: The account being accessed does not support http.
	at com.microsoft.azure.storage.StorageException.translateFromHttpStatus(StorageException.java:175)
	at com.microsoft.azure.storage.StorageException.translateException(StorageException.java:94)
	at com.microsoft.azure.storage.core.StorageRequest.materializeException(StorageRequest.java:315)
	at com.microsoft.azure.storage.core.ExecutionEngine.executeWithRetry(ExecutionEngine.java:175)
	at com.microsoft.azure.storage.blob.CloudBlobContainer.exists(CloudBlobContainer.java:769)
	at com.microsoft.azure.storage.blob.CloudBlobContainer.createIfNotExists(CloudBlobContainer.java:379)
	at com.microsoft.azure.storage.blob.CloudBlobContainer.createIfNotExists(CloudBlobContainer.java:348)
	at com.microsoft.azure.eventprocessorhost.AzureStorageCheckpointLeaseManager.lambda$createLeaseStoreIfNotExistsInternal$8(AzureStorageCheckpointLeaseManager.java:328)
	at java.util.concurrent.CompletableFuture$AsyncRun.run(Unknown Source)
	at java.util.concurrent.Executors$RunnableAdapter.call(Unknown Source)
	at java.util.concurrent.FutureTask.run(Unknown Source)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$201(Unknown Source)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(Unknown Source)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
	at java.lang.Thread.run(Unknown Source)`

## Proposed Solution
* Is there a property that I don't know about that would allow me to make HTTP**S** requests instead? If so please share more information. If not, it would be nice to have HTTP**S** as default protocol in a version that's basically 1.0.0.M2 but with HTTPS as default protocol instead of HTTP. It is not possible for me to migrate to spring-cloud-azure-eventhub-stream-binder 1.1.X due to heavy dependency incompatibilities.

## Source Code
The class `com.microsoft.azure.spring.cloud.context.core.impl.StorageConnectionStringProvider.ConnectionStringBuilder` in the version 1.0.0.M2 has the following constant `private static final String HTTP_PROTOCOL = "http";` set, which causes the issue. Simply replacing its value with `"https"` instead, would solve this issue.

## Your Environment
* Version used: spring-cloud-azure-eventhub-stream-binder 1.0.0.M2
* Operating System and version (desktop or mobile): Windows 10 64-bit
* SDK version: Java(TM) SE Runtime Environment (build 1.8.0_74-b02)

dsibilio:
## Description
Grant a configuration point regarding the default HTTP protocol to be used for Storage Account connection when eg. checkpointing.
A new property has been added: **spring.cloud.azure.storage.secureTransfer** (not mandatory, **default value true**) to manage this configuration point.

### unset
1. Configure a storage account with Secure Transfer enabled
2. Let the spring.cloud.azure.storage.secureTransfer be unset
3. Use the Azure EventHubs Stream Binder specifying the forementioned Storage Account (_spring.cloud.azure.eventhub.checkpoint-storage-account=my-storage-account_)
4. There should be no errors when connecting/checkpointing

### true
1. Configure a storage account with Secure Transfer enabled
2. Set the spring.cloud.azure.storage.secureTransfer to **true**
3. Use the Azure EventHubs Stream Binder specifying the forementioned Storage Account (_spring.cloud.azure.eventhub.checkpoint-storage-account=my-storage-account_)
4. There should be no errors when connecting/checkpointing

### false
1. Configure a storage account with Secure Transfer enabled
2. Set the spring.cloud.azure.storage.secureTransfer to **false**
3. Use the Azure EventHubs Stream Binder specifying the forementioned Storage Account (_spring.cloud.azure.eventhub.checkpoint-storage-account=my-storage-account_)
4. There should be the following error when bootstrapping the application: `com.microsoft.azure.storage.StorageException: The account being accessed does not support http`

### DIFF
--- a/spring-cloud-azure-context/src/main/java/com/microsoft/azure/spring/cloud/context/core/config/AzureProperties.java
+++ b/spring-cloud-azure-context/src/main/java/com/microsoft/azure/spring/cloud/context/core/config/AzureProperties.java
@@ -8,7 +8,6 @@ package com.microsoft.azure.spring.cloud.context.core.config;
 
 import com.microsoft.azure.spring.cloud.context.core.api.CredentialSupplier;
 import com.microsoft.azure.spring.cloud.context.core.api.Environment;
-
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.util.Assert;
 import org.springframework.validation.annotation.Validated;
@@ -28,8 +27,6 @@ public class AzureProperties implements CredentialSupplier {
     private String region;
 
     private boolean autoCreateResources = false;
-
-    private String defaultHttpProtocol = "https";
 
     @PostConstruct
     private void validate() {
@@ -78,13 +75,5 @@ public class AzureProperties implements CredentialSupplier {
 
     public void setAutoCreateResources(boolean autoCreateResources) {
         this.autoCreateResources = autoCreateResources;
-    }
-
-    public String getDefaultHttpProtocol() {
-        return defaultHttpProtocol;
-    }
-
-    public void setDefaultHttpProtocol(String defaultHttpProtocol) {
-        this.defaultHttpProtocol = defaultHttpProtocol;
     }
 }
--- a/spring-cloud-azure-context/src/main/java/com/microsoft/azure/spring/cloud/context/core/config/AzureProperties.java
+++ b/spring-cloud-azure-context/src/main/java/com/microsoft/azure/spring/cloud/context/core/config/AzureProperties.java
@@ -8,6 +8,7 @@ package com.microsoft.azure.spring.cloud.context.core.config;
 
 import com.microsoft.azure.spring.cloud.context.core.api.CredentialSupplier;
 import com.microsoft.azure.spring.cloud.context.core.api.Environment;
+
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.util.Assert;
 import org.springframework.validation.annotation.Validated;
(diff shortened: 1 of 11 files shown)


spring-cloud-azure-context/src/main/java/com/microsoft/azure/spring/cloud/context/core/config/AzureProperties.java
```diff
@@ -28,6 +29,8 @@
 
     private boolean autoCreateResources = false;
 
+    private boolean secureTransfer = true;
```
warrenzhu25: Could we move this property into storage properties? It seems it only applies to storage account.

dsibilio: Good point. Pushin the changes right now.

PR #543, opened 2019-08-02
fix #542 : Reregister event hub consumer after unregister caused IllegalStateException

### LINKED ISSUE #542: Reregister event hub consumer after unregister caused IllegalStateException
# Steps to reproduce:
1. when the microservice starts all binders go up and … all is ok
2. after several tasks I decide to stop the binders and … all is ok
3. after some while I decide to restart the binders….I have a problem

# Stacktrace as below:
```
reactor.core.Exceptions$ErrorCallbackNotImplemented: java.lang.IllegalStateException: Register cannot be called on an EventProcessorHost after unregister. Please create a new EventProcessorHost instance.
Caused by: java.lang.IllegalStateException: Register cannot be called on an EventProcessorHost after unregister. Please create a new EventProcessorHost instance.
at com.microsoft.azure.eventprocessorhost.EventProcessorHost.registerEventProcessorFactory(EventProcessorHost.java:461)
at com.microsoft.azure.spring.integration.eventhub.impl.AbstractEventHubTemplate.register(AbstractEventHubTemplate.java:103)
at com.microsoft.azure.spring.integration.eventhub.impl.EventHubTemplate.subscribe(EventHubTemplate.java:40)
at com.microsoft.azure.spring.integration.core.api.SubscribeByGroupOperation.subscribe(SubscribeByGroupOperation.java:30)
at com.microsoft.azure.spring.integration.core.AbstractInboundChannelAdapter.doStart(AbstractInboundChannelAdapter.java:38)
at org.springframework.integration.endpoint.AbstractEndpoint.start(AbstractEndpoint.java:156)
at org.springframework.cloud.stream.binder.DefaultBinding.start(DefaultBinding.java:116)
at com.wba.oneit.oneleo.config.BinderStartStopService.lambda$start$0(BinderStartStopService.java:41)
at java.util.concurrent.ConcurrentHashMap$EntrySetView.forEach(ConcurrentHashMap.java:4795)
at com.wba.oneit.oneleo.config.BinderStartStopService.start(BinderStartStopService.java:39)
```
# The root cause is inside:
```
protected void register(String name, String consumerGroup, EventHubProcessor eventProcessor) {
        EventProcessorHost host = this.clientFactory.getOrCreateEventProcessorHost(name, consumerGroup);
        host.registerEventProcessorFactory(context -> eventProcessor, buildEventProcessorOptions(startPosition));
    }

    protected void unregister(String name, String consumerGroup) {
        this.clientFactory.getOrCreateEventProcessorHost(name, consumerGroup).unregisterEventProcessor()
                          .whenComplete((s, t) -> {
                              if (t != null) {
                                  log.warn(String.format("Failed to unregister consumer '%s' with group '%s'", name,
                                          consumerGroup), t);
                              }
                          });
```
where this.clientFactory.getOrCreateEventProcessorHost is called in both register and unregister cases. 

But when I do the mentioned sequence of operations (register , unregister,  register)  I have an error during the second register phase because the EventProcessorHost retrieved from the map 
is registered and it has valorised variables that are null during the first register phase .

# How to fix

It seems that we should remove `EventProcessorHost` after unregister, then creating new one after reregister.

ivancerrato:
## Description
fix #542 : Reregister event hub consumer after unregister caused IllegalStateException

## Related PRs

branch | PR

## Todos

Note:

Within  com.microsoft.azure.spring.integration.eventhub.impl.AbstractEventHubTemplate.unregister(String, String)

the method com.microsoft.azure.spring.integration.eventhub.api.EventHubClientFactory.getOrCreateEventProcessorHost(String, String) is called.

This implies that an existing or new instance of EventProcessorHost is retrieved.
But it seems very strange creating a new instance of EventProcessorHost during the **unregister** phase.

## Steps to Test

1. run  microservice using EventHub that starts all binders
2. using microservice, make all binders to go down 
3. after some while, using microservice, make all binders to go up

### DIFF
--- a/spring-cloud-azure-context/src/main/java/com/microsoft/azure/spring/cloud/context/core/util/Memoizer.java
+++ b/spring-cloud-azure-context/src/main/java/com/microsoft/azure/spring/cloud/context/core/util/Memoizer.java
@@ -34,4 +34,10 @@ public class Memoizer {
     public static <T, U, R> BiFunction<T, U, R> memoize(Map<Tuple<T, U>, R> map, BiFunction<T, U, R> biFunction) {
         return (t, u) -> map.computeIfAbsent(Tuple.of(t, u), (k) -> biFunction.apply(k.getFirst(), k.getSecond()));
     }
+    
+    public static <T, U, R> BiFunction<T, U, R> memoizeCompute
+      (Map<Tuple<T, U>, R> map, BiFunction<T, U, R> biFunction) {
+        return (t, u) -> map.compute(Tuple.of(t, u), (k, v) -> biFunction.apply(k.getFirst(), k.getSecond()));
+    }
+    
 }
--- a/spring-integration-azure/spring-integration-eventhubs/src/main/java/com/microsoft/azure/spring/integration/eventhub/factory/DefaultEventHubClientFactory.java
+++ b/spring-integration-azure/spring-integration-eventhubs/src/main/java/com/microsoft/azure/spring/integration/eventhub/factory/DefaultEventHubClientFactory.java
@@ -16,6 +16,9 @@ import com.microsoft.azure.spring.cloud.context.core.util.Tuple;
 import com.microsoft.azure.spring.integration.eventhub.api.EventHubClientFactory;
 import com.microsoft.azure.spring.integration.eventhub.impl.EventHubRuntimeException;
 import com.microsoft.azure.spring.integration.eventhub.util.HostnameHelper;
+
+import reactor.util.function.Tuples;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.DisposableBean;
@@ -56,7 +59,9 @@ public class DefaultEventHubClientFactory implements EventHubClientFactory, Disp
             Memoizer.memoize(clientsByName, this::createEventHubClient);
     private final BiFunction<String, String, EventProcessorHost> processorHostCreator =
             Memoizer.memoize(processorHostMap, this::createEventProcessorHost);
-
+    private final BiFunction<String, String, EventProcessorHost> processorHostCreatorForce =
+            Memoizer.memoizeCompute(processorHostMap, this::createEventProcessorHost);
+    
     public DefaultEventHubClientFactory(@NonNull EventHubConnectionStringProvider connectionStringProvider,
             String checkpointConnectionString) {
         Assert.hasText(checkpointConnectionString, "checkpointConnectionString can't be null or empty");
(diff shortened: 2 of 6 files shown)


spring-integration-azure/spring-integration-eventhubs/src/main/java/com/microsoft/azure/spring/integration/eventhub/factory/DefaultEventHubClientFactory.java
```diff
@@ -83,9 +88,12 @@ private PartitionSender createPartitionSender(EventHubClient client, String part
     }
 
     private EventProcessorHost createEventProcessorHost(String name, String consumerGroup) {
-        return new EventProcessorHost(EventProcessorHost.createHostName(HostnameHelper.getHostname()), name,
+        EventProcessorHost  eventProcessorHost  = null;
```
warrenzhu25: Any specific reason to split this?

spring-cloud-azure-context/src/main/java/com/microsoft/azure/spring/cloud/context/core/util/Memoizer.java
```diff
@@ -34,4 +34,10 @@
     public static <T, U, R> BiFunction<T, U, R> memoize(Map<Tuple<T, U>, R> map, BiFunction<T, U, R> biFunction) {
         return (t, u) -> map.computeIfAbsent(Tuple.of(t, u), (k) -> biFunction.apply(k.getFirst(), k.getSecond()));
     }
+    
+    public static <T, U, R> BiFunction<T, U, R> memoizeCompute
```
warrenzhu25: Try to understand your change. You try to create a new instance of EventProcessorHost even when old one existed. How about just changing the logic of unregister?

1. Add get method for this, rather than getOrCreate
2. Inside unregister, if existed, remove from cache, otherwise, do nothing.

ivancerrato: Dear Friend...
maybe only you have a deep knowledge of the code(functional behaviour)...I can suggest a non-invasive modification to allow the code works without crash. But, as already said, there are a strange behaviour that, frankly speaking, I didn't understand. During the unregister (into com.microsoft.azure.spring.integration.eventhub.impl.AbstractEventHubTemplate) even when a EventProcessorHost is null ...a new one is being created. So I cannot modify a great chunk of your code without having a whole picture of the functional behaviour. And moreover I cannot modify a substantial part of code without ask to you more info....it's common sense.

warrenzhu25: Thanks anyway. May I close this? I'll handle the fix.

chenghaoharvey: > Try to understand your change. You try to create a new instance of EventProcessorHost even when old one existed. How about just changing the logic of unregister?
> 
> 1. Add get method for this, rather than getOrCreate
> 2. Inside unregister, if existed, remove from cache, otherwise, do nothing.

hi , i try to fix this bug. May i confirm that
1. Add remove method in EventHubClientFactory to remove EventProcessorHost from cache
2. As showed below, Inside unregister, if existed, remove from cache, otherwise, do nothing.
`protected void unregister(String name, String consumerGroup) {
        this.clientFactory.getOrCreateEventProcessorHost(name, consumerGroup).unregisterEventProcessor()
                          .whenComplete((s, t) -> {
                              if (t != null) {
                                  log.warn(String.format("Failed to unregister consumer '%s' with group '%s'", name,
                                          consumerGroup), t);
                              }else{
                                  this.clientFactory.removeEventProcessorHost(name, consumerGroup);
                              }
                          });
    }`

dsibilio: Hi @warrenzhu25 , @chenghaoharvey .
You can check my take - following @warrenzhu25 feedback - on this bug @ #546

PR #545, opened 2019-08-06
Feature management samples and docs

mrm9084:
## Description
Adds Documentation and Examples of Feature Management


spring-cloud-azure-feature-management/README.md
```diff
@@ -0,0 +1,129 @@
+# Feature Management
+Feature flags provide a way for Spring Boot applications to turn features on or off dynamically. Developers can use feature flags in simple use cases like conditional statement to more advanced scenarios like conditionally adding routes. Feature Flags are not dependent of any spring-cloud-azure dependencies, but may be used in conjunction with spring-cloud-azure-appconfiguration-config.
+
+Here are some of the benefits of using this library:
+* A common convention for feature management
+* Low barrier-to-entry
+  * Supports application.yml file feature flag setup
```
sophiaso: I think it should be `bootstrap.yml`?

spring-cloud-azure-feature-management/README.md
```diff
@@ -0,0 +1,129 @@
...
+@ResponseBody
+public String oldEndpoint() {
+    ...
+}
+```
+
+## Implementing a Feature Filter
+Creating a feature filter provides a way to enable features bassed on criteria that you define. To implement a feature filter, the `FeatureFilter` interface must be implemented. `FeatureFilter` has a single method `evaluate`. When a feature specifies that it can be enabled for a feature filter, the `evaluate` method is called. method is called. If `evaluate` returns `true` it means the feature should be enabled.
+
+Feature filters are found by being defined as `@Component` where there name matches the expected filter defined in the configuration.
```
jimmyca15: You may want to provide a code snippet so readers can see exactly how the name maps, what parameters they should be expect, what annotations they need to add etc. all in one place.

spring-cloud-azure-feature-management/README.md
```diff
@@ -0,0 +1,129 @@
...
+
+## Request Based Features/Snapshot
+There are scenarios which require the state of a feature to remain consistent during the lifetime of a request. The values returned from the standard `FeatureManager` may change if the configuration source which it is pulling from is updated during the request. This can be prevented by using `FeatureManagerSnapshot` and `@FeatureOn( snapshot = true )`. `FeatureManagerSnapshot` can be retrieved in the same manner as `FeatureManager`. `FeatureManagerSnapshot` calls `FeatureManager`, but it caches the first evaluated state of a feature during a request and will return the same state of a feature during its lifetime.
```
jimmyca15: Other than the room for a feature filter example, everything looks good to me.

PR #553, opened 2019-08-28
fix user-assigned managed identity for app service, and update sample…

saragluna:
… and doc

## Description
- MSI support for app service using user-assigned identity.
- Update sample
- Update doc

### DIFF
--- a/spring-cloud-azure-dependencies/pom.xml
+++ b/spring-cloud-azure-dependencies/pom.xml
@@ -26,16 +26,15 @@
         <commons.io.version>2.6</commons.io.version>
         <hibernate.validator.version>6.0.12.Final</hibernate.validator.version>
         <google.guava.version>20.0</google.guava.version>
-        <azure.sdk.version>1.14.0</azure.sdk.version>
+        <azure.sdk.version>1.23.0</azure.sdk.version>
         <azure.eventhubs.eph.version>2.4.0</azure.eventhubs.eph.version>
         <azure.keyvault.version>1.0.0</azure.keyvault.version>
-        <azure.client.version>1.6.0</azure.client.version>
         <azure.servicebus.version>1.2.8</azure.servicebus.version>
         <javax.jms-api.version>2.0.1</javax.jms-api.version>
         <spring-jms.version>5.1.8.RELEASE</spring-jms.version>
         <qpid-jms-client.version>0.43.0</qpid-jms-client.version>
-
         <spring.cloud.azure.version>1.1.0-SNAPSHOT</spring.cloud.azure.version>
+        <okhttp3.version>3.11.0</okhttp3.version>
     </properties>
 
     <dependencyManagement>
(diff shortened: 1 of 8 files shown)


spring-cloud-azure-samples/eventhubs-binder-sample/pom.xml
```diff
@@ -23,4 +23,15 @@
         </dependency>
     </dependencies>
 
+    <build>
```
warrenzhu25: Why add this in sample?

spring-cloud-azure-samples/eventhubs-binder-sample/src/main/java/com/example/SourceExample.java
```diff
@@ -30,4 +31,9 @@ public String postMessage(@RequestParam String message) {
         this.source.output().send(new GenericMessage<>(message));
         return message;
     }
+
```
warrenzhu25: What's this?

spring-cloud-azure-samples/eventhubs-binder-sample/pom.xml
saragluna: User can use this plugin to deply jar to app service.

spring-cloud-azure-samples/eventhubs-binder-sample/README.adoc
```diff
@@ -112,6 +112,50 @@ spring.cloud.azure.eventhub.checkpoint-storage-account=[checkpoint-storage-accou
 spring.cloud.azure.eventhub.namespace=[eventhub-namespace]
 ....
 
+=== Steps for application to run in App Service
```
warrenzhu25: Could you remove the duplicate part with vm?

spring-cloud-azure-samples/eventhubs-binder-sample/src/main/java/com/example/SourceExample.java
saragluna: easy way to check wheter the app starts successfully in App Service, instead of using curl, user can request this via browser

spring-cloud-azure-samples/eventhubs-binder-sample/README.adoc
saragluna: Sure, then the two parts will mix together

spring-cloud-azure-dependencies/pom.xml
```diff
@@ -26,16 +26,15 @@
         <commons.io.version>2.6</commons.io.version>
         <hibernate.validator.version>6.0.12.Final</hibernate.validator.version>
         <google.guava.version>20.0</google.guava.version>
-        <azure.sdk.version>1.14.0</azure.sdk.version>
+        <azure.sdk.version>1.23.0</azure.sdk.version>
```
warrenzhu25: Is old version blocking this feature? If you update this, you need to do full integration tests of all modules.

saragluna: yes, old version doesn't support clientid when user-assigned identity, and according to their document:
>
clientid | Query | (Optional) The ID of the user-assigned identity to be used. If omitted, the system-assigned identity is used.
-- | -- | --

saragluna: sure, I will run full integration test

saragluna:
Run full integration test on local machine.

PR #554, opened 2019-08-28
App Configuration - Added Minimum Refresh Time, Switch to FilterNotFoundException, Removed Enabled from Feature Mangement

mrm9084:
A few minor changes to App Configuration and Feature Management. Set a minimum watch time between App Configuration refreshes to 1 second. Added a FilterNotFoundException to send instead of NoSuchBeanDefinitionException. We decided to remove enabled from Feature Management as it added confusion.


spring-cloud-azure-feature-management/src/main/java/com/microsoft/azure/spring/cloud/feature/manager/FilterNotFoundException.java
```diff
@@ -0,0 +1,43 @@
...
+        super(message, cause);
+        this.filter = filter;
+    }
+
+    @Override
+    public String getMessage() {
+        if (filter == null) {
+            return getCause().getMessage();
+        }
+        return getCause().getMessage() + ", " + filter.toString();
```
sophiaso: nit: is it showing the `filter.toString()` only in the message a little hard for the user to understand the error message?

PR #562, opened 2019-09-18
App configuration ga updates

mrm9084:
Adds

- APIVersion on requests
- Fixes Issue in Auth Header where commas were used instead of &
- Fixes some Javadoc Errors
- Fixes Watch when Prefix or Profiles are used
- Fixes Bug in Feature Management when null values is passed
- Updates README to say Contributor is needed for Managed Identity


spring-cloud-azure-appconfiguration-config/src/main/java/com/microsoft/azure/spring/cloud/config/AzureCloudConfigWatch.java
```diff
@@ -186,11 +186,10 @@ public String getMessage() {
      * @return the full name of the key mapping to the configuration store
      */
     private String watchedKeyNames(ConfigStore store, Map<String, List<String>> storeContextsMap) {
-        String prefix = store.getPrefix();
         String watchedKey = store.getWatchedKey().trim();
         List<String> contexts = storeContextsMap.get(store.getName());
 
-        String watchedKeys = contexts.stream().map(ctx -> genKey(prefix, ctx, watchedKey))
+        String watchedKeys = contexts.stream().map(ctx -> genKey(ctx, watchedKey))
```
saragluna: Why remove the prefix？

mrm9084: The issues is that ctx already has the prefix attached to it, so it resulted in /<prefix>/<prefix>/application>/

PR #568, opened 2019-10-02
Compute settings

mrm9084:
Adds support for computing values in the bootstrap config file.

i.e.
spring.cloud.azure.appconfiguration.watch.enabled=${enable.watch}

where enabled.watch can be defined in App Configuration.


spring-cloud-azure-appconfiguration-config/src/main/java/com/microsoft/azure/spring/cloud/config/AzureConfigBootstrapConfiguration.java
```diff
@@ -46,7 +49,7 @@
 
     @Bean
     public ConnectionStringPool initConnectionString(AzureCloudConfigProperties properties,
-                                                     AzureTokenCredentials credentials) {
+            AzureTokenCredentials credentials, AbstractEnvironment env) {
```
x9xu: env seems not to be used in this method

PR #569, opened 2019-10-08
Removes Property Cache

mrm9084:
Removes Local Caching, moves content-types to Constants file, switches exampe to use Enviroment variable


spring-cloud-azure-appconfiguration-config/src/main/java/com/microsoft/azure/spring/cloud/config/AzureCloudConfigWatch.java
```diff
@@ -76,15 +77,30 @@ public void setApplicationEventPublisher(ApplicationEventPublisher applicationEv
      */
     public void refreshConfigurations() {
         if (this.running.compareAndSet(false, true)) {
+            Date date = new Date();
+            Boolean refreshed = false;
             for (ConfigStore configStore : configStores) {
-                if (propertyCache.findNonCachedKeys(delay, configStore.getName()).size() > 0) {
+                Boolean needsRefresh = false;
+                if (!lastUpdated.containsKey(configStore.getName())) {
```
x9xu: Wondering where this lastUpdated gets updated with refreshed configuration stores

PR #573, opened 2019-10-17
Switch App Configuration to use the azure sdk

mrm9084:


spring-cloud-azure-appconfiguration-config/src/main/java/com/microsoft/azure/spring/cloud/config/AzureCloudConfigWatch.java
```diff
@@ -111,16 +115,22 @@ public void refreshConfigurations() {
...
-                .withLabels(store.getLabels()).withFields(QueryField.ETAG).withRange(0, 0);
-
-        List<KeyValueItem> keyValueItems = configOperations.getRevisions(store.getName(), options);
+        SettingSelector settingSelector = new SettingSelector().setKeys(watchedKeyNames).setLabels(store.getLabels())
+                .setRange(new Range(0, 0));
+
+        List<ConfigurationSetting> items = null;
+        try {
+            items = clientStore.listSettingRevisons(settingSelector, store.getName());
+        } catch (ServerException e) {
```
jialindai: I am assuming the underlying protocol is http?
If that's the case, it's strange to use one rmi exception.

spring-cloud-azure-appconfiguration-config/src/main/java/com/microsoft/azure/spring/cloud/config/AzureConfigPropertySource.java
```diff
@@ -135,15 +147,12 @@ public FeatureSet initProperties(FeatureSet featureSet) throws IOException {
...
 
             // Parsing Key Vault Reference for URI
             try {
                 JsonNode kvReference = mapper.readTree(value);
-                stringUri = kvReference.at("/uri").asText();
-                uri = new URI(stringUri);
+                uri = new URI(kvReference.at("/uri").asText());
             } catch (URISyntaxException e) {
                 if (azureProperties.isFailFast()) {
                     LOGGER.error("Error Processing Key Vault Entry URI.", e);
```
jialindai: Seems should throw exception in the fail fast condition?

spring-cloud-azure-appconfiguration-config/src/main/java/com/microsoft/azure/spring/cloud/config/AzureConfigPropertySourceLocator.java
```diff
@@ -176,22 +181,28 @@ private String propWithProfile(String context, String profile) {
...
+                if (initFeatures) {
+                    propertySource.initFeatures(featureSet);
+                }
+                sourceList.add(propertySource);
+                putStoreContext(store.getName(), context, storeContextsMap);
             }
-            sourceList.add(propertySource);
-            putStoreContext(store.getName(), context, storeContextsMap);
+        } catch (Exception e) {
+            delayException();
```
jialindai: What's the purpose of delayException?

spring-cloud-azure-appconfiguration-config/src/main/java/com/microsoft/azure/spring/cloud/config/stores/ClientStore.java
```diff
@@ -0,0 +1,144 @@
...
+                    throw new IllegalArgumentException("Connections String can't be empty or null");
+                }
+
+                ConfigurationClient client = builder.buildClient();
+
+                configClients.put(store, client);
+            } catch (IllegalArgumentException e) {
+                LOGGER.error("Failed to load Config Store.");
+                if (properties.isFailFast()) {
+                    ReflectionUtils.rethrowRuntimeException(new Exception("Failed to load Config Store.", e));
```
jialindai: ReflectionUtils can be avoided

spring-cloud-azure-appconfiguration-config/src/main/java/com/microsoft/azure/spring/cloud/config/AzureConfigPropertySourceLocator.java
mrm9084: There was a thing seen in applications that automatically restart when they fail. In these cases a large number of requests to App Configuration can be made in a very short period of time. The idea behind delayException is to delay a restart if App Configuration fails to load for any reason, this is done by setting a minimum up time before any error is thrown.

spring-cloud-azure-appconfiguration-config/src/main/java/com/microsoft/azure/spring/cloud/config/AzureCloudConfigWatch.java
mrm9084: This shouldn't be there. This exception isn't thrown, removing.

PR #601, opened 2020-01-08
Feature management async

mrm9084:
- Updated the Feature Management Libraries to be Nonblocking Async with Reactor.
- Added two pre made filters PercentageFilter and TimeWindow
- Updated so the config file now matches the DotNet Feature Management config file format


spring-cloud-azure-feature-management/src/main/java/com/microsoft/azure/spring/cloud/feature/manager/feature/filters/PercentageFilter.java
```diff
@@ -0,0 +1,51 @@
...
+
+    /**
+     * Performs a percentage based evaluation to determine whether a feature is enabled.
+     * @param context The feature evaluation context.
+     * @return True if the feature is enabled, false otherwise.
+     */
+    @Override
+    public boolean evaluate(FeatureFilterEvaluationContext context) {
+        try {
+            Thread.sleep(5000);
```
x9xu: Wondering why it gonna sleep for 5 sec here?

mrm9084: It must have been from a test and missed. Will remove.

spring-cloud-azure-feature-management/src/main/java/com/microsoft/azure/spring/cloud/feature/manager/feature/filters/TimeWindowFilter.java
```diff
@@ -0,0 +1,56 @@
...
+        ZonedDateTime now = ZonedDateTime.now();
+
+        if (!StringUtils.hasText(start) && !StringUtils.hasText(end)) {
+            LOGGER.warn("The {} feature filter is not valid for feature {}. It must specify either {}, {}, or both.",
+                    this.getClass().getSimpleName(), context.getName(), TIME_WINDOW_FILTER_SETTING_START,
+                    TIME_WINDOW_FILTER_SETTING_END);
+            return false;
+        }
+
+        ZonedDateTime startTime = ZonedDateTime.parse(start, DateTimeFormatter.RFC_1123_DATE_TIME);
```
x9xu: We assume that the start and end are always not null here? If only of them is not specified in context, what will be returned at line 37 or 38?

https://docs.oracle.com/javase/8/docs/api/java/time/ZonedDateTime.html#parse-java.lang.CharSequence-java.time.format.DateTimeFormatter-

mrm9084: I thought I had tested this given the return, but it doesn't seem to work. Added a check.

PR #605, opened 2020-01-09
App Config - Refresh Async

mrm9084:
* AzureCloudConfigRefresh is now non-blocking Async using Reactor.
* Bug fix, multiple refreshes no longer occur when a system that is watching multiple stores where two or more stores are updated before the local values are updated.
* Bug fix, Sentinels are update before refresh completes, resulting in a failed refresh not being reattempted.


spring-cloud-azure-appconfiguration-config/src/main/java/com/microsoft/azure/spring/cloud/config/AzureCloudConfigRefresh.java
```diff
@@ -74,31 +67,56 @@ public void setApplicationEventPublisher(ApplicationEventPublisher applicationEv
...
-                    if (refreshed) {
-                        break;
+    /**
+     * Goes through each config store and checks if any of its keys need to be refreshed.
+     * If any store has a value that needs to be updated a refresh event is called after
+     * every store is checked.
+     * @return If a refresh event is called.
+     */
+    private boolean refreshStores() {
+        boolean needsRefresh = false;
```
x9xu: Rename it with a another word. In the java doc, it tells the user that If a refresh event is called, but needsRefresh makes users feel that this method will tell them if they need to do it or not.

PR #607, opened 2020-01-13
App Config Feature Fixes

mrm9084:
## Description
Fixes:

- Bug in App Config Test where a .get() was missed on an Async method so the operation didn't always complete before being checked.
- Added extra check on Refresh to check for invalid state and reset.
- Fixed issue in Feature Management where local config features and App Configuration features both didn't work with it.
- Updated App Configuration starter README to be match recent changes

## Related PRs
PR #? - Same update, but to 1.1.x

branch | PR
------ | ------
other_pr_production | [link]()
other_pr_master | [link]()

## Todos
- [x] Tests
- [x] Documentation

## Steps to Test

* firstCallShouldPublishEvent should no longer fail on a slow system.
* Both the feature sample config file and the matching configurations in App Configuration will both work the same.


spring-cloud-azure-starters/spring-cloud-starter-azure-appconfiguration-config/README.md
```diff
@@ -108,33 +108,32 @@ spring.cloud.azure.appconfiguration.fail-fast=false
 
 [Managed identity](https://docs.microsoft.com/azure/active-directory/managed-identities-azure-resources/overview) allows application to access [Azure Active Directory][azure_active_directory] protected resource on [Azure][azure].
 
-In this library, [Azure Identity SDK][azure_identity_sdk] is used to access Azure App Configuration and optionally Azure Key Vault, for secrets. Only one method of authentication can be set at one time.
+In this library, [Azure Identity SDK][azure_identity_sdk] is used to access Azure App Configuration and optionally Azure Key Vault, for secrets. Only one method of authentication can be set at one time. When not using the AppConfigCredentialProvider and/or KeyVaultCredentialProvider the same authentication method is used for both App Configuration and Key Vault.
```
zhenlan: **Note:** By default ~~When not using the AppConfigCredentialProvider and/or KeyVaultCredentialProvider~~ the same ~~authentication method~~ managed identity is used to access both App Configuration and Key Vault.

mrm9084: I am not sure this is the correct way to word this either. The "default" way of doing Managed Identity is done by setting the endpoint, for System Assigned, or the endpoint and a client id for user assigned. I was trying to say in that case both will be using the same credential.

PR #621, opened 2020-02-05
App Config Readme Updates

mrm9084:
## Description
Updates the App Configuration and Feature Management Readme files with recent changes.


spring-cloud-azure-starters/spring-cloud-starter-azure-appconfiguration-config/README.md
```diff
@@ -32,15 +32,15 @@ or
```
zhenlan: Line 1: Can we change the title of the document to something like **Azure App Configuration for Spring Cloud**?

Line 13 & 23: We should add a note when to use which dependency (eg., use the -web one in a web application...)

spring-cloud-azure-starters/spring-cloud-starter-azure-appconfiguration-config/README.md
```diff
@@ -60,11 +60,11 @@ spring.cloud.azure.appconfiguration.managed-identity.client-id | Client id of th
 
 Name | Description | Required | Default
 ---|---|---|---
-spring.cloud.azure.appconfiguration.stores[0].endpoint | Endpoint of the configuration store, required when `connection-string` is empty. If `connection-string` is empty and application is deployed on Azure VM or App Service with managed identity enabled, will try to load `connection-string` from Azure Resource Management. | Conditional | null
+spring.cloud.azure.appconfiguration.stores[0].endpoint | Endpoint of the configuration store, required when `connection-string` is empty. If `connection-string` is empty alternate connection methods will be attempted based on configuration, such as System Assigned Identity, or User Assigned Identity if the client-id value is set. | Conditional | null
```
zhenlan: I think we should reword the description and focus on explaining when the endpoint should be used, for example,

*When the endpoint of an App Configuration store is specified, a managed identity or a token credential provided using `AppConfigCredentialProvider` will be used to connect to the App Configuration service. Errors (what error?) will be thrown if the endpoint and connection-string are specified at the same time.*

We don't have to explain the differences between system-assigned managed identity and user-assigned managed identity here.

spring-cloud-azure-starters/spring-cloud-starter-azure-appconfiguration-config/README.md
```diff
@@ -60,11 +60,11 @@ spring.cloud.azure.appconfiguration.managed-identity.client-id | Client id of th
...
 spring.cloud.azure.appconfiguration.stores[0].prefix | The prefix of the key name in the configuration store, e.g., /my-prefix/application/key.name | No |  null
-spring.cloud.azure.appconfiguration.stores[0].connection-string | Required when `name` is empty, otherwise, can be loaded automatically on Azure Virtual Machine or App Service | Conditional | null
+spring.cloud.azure.appconfiguration.stores[0].connection-string | Required when `endpoint` is empty, otherwise, can be loaded automatically on Azure Virtual Machine or App Service | Conditional | null
```
zhenlan: Similarly, reword this and focus on explaining when the connection-string should be used, for example,

*When the connection-string of an App Configuration store is specified, HMAC authentication will be used to connect to the App Configuration service. Errors (what error?) will be thrown if the endpoint and connection-string are specified at the same time.*

spring-cloud-azure-starters/spring-cloud-starter-azure-appconfiguration-config/README.md
```diff
@@ -60,11 +60,11 @@ spring.cloud.azure.appconfiguration.managed-identity.client-id | Client id of th
...
 spring.cloud.azure.appconfiguration.stores[0].label | Comma separated list of label values, by default will query empty labeled value. If you want to specify *empty*(null) label explicitly, use `%00`, e.g., spring.cloud.azure.appconfiguration.stores[0].label=%00,v0 | No |  null
-spring.cloud.azure.appconfiguration.stores[0].fail-fast | Whether throw RuntimeException or not when exception occurs. If an exception does occur when false the store is skipped. | No |  true
+spring.cloud.azure.appconfiguration.stores[0].fail-fast | Whether throw RuntimeException or not when exception occurs. If an exception does occur during startup when set to false the store is skipped. | No |  true
```
zhenlan: We should be clear about this affects start-up only. For example,

*Whether throw `RuntimeException` or not when fail to read App Configuration during application start-up. If an exception does occur during startup when set to false the store is skipped.*

spring-cloud-azure-starters/spring-cloud-starter-azure-appconfiguration-config/README.md
mrm9084: For Line 13 & 23 I was thinking the same thing, but couldn't figure out how to word it. In the sections below I mention when to use each one.

PR #630, opened 2020-02-21
AppConfig 1.1.2 Updates

mrm9084:
## Description
App Config Updates for 1.1.2


pom.xml
```diff
@@ -53,6 +53,7 @@
 
     <modules>
         <module>spring-cloud-azure-appconfiguration-config</module>
+		<module>spring-cloud-azure-appconfiguration-config-web</module>
```
saragluna: Wrong indention.

spring-cloud-azure-appconfiguration-config-web/src/main/java/com/microsoft/azure/spring/cloud/config/web/AppConfigurationWebAutoConfiguration.java
```diff
@@ -0,0 +1,27 @@
...
+
+@Configuration
+public class AppConfigurationWebAutoConfiguration {
+
+    @Configuration
+    @ConditionalOnClass(RefreshEndpoint.class)
+    static class AppConfigurationWatchAutoConfiguration {
+
+        @Bean
+        public ConfigListener configListener(AppConfigurationRefresh appConfigurationRefresh) {
```
saragluna: If the configListener bean depends on RefreshEndpoint.class, should we put @Component annotation on the ConfigListner class?

mrm9084: @saragluna ConfigListener is already marked as a component. See line 16 of the ConfigListener.

saragluna: Sorry for not making myself clear. I mean we tend to configure this bean on some condition, however, putting a @component annotation will always initiate that bean.  Is this what we want?

mrm9084: @saragluna Yes, at least for now. if you look at the package as a whole it only has two classes right now, which are this configuration class and the listener. This might not always be the case but at least for now the listener should always be created when using this package.

PR #636, opened 2020-03-05
Event Hub Track2 Library Migration

saragluna:
## Description
Migrate to Event Hub track2 library.


spring-cloud-azure-autoconfigure/src/main/java/com/microsoft/azure/spring/cloud/autoconfigure/eventhub/AzureEventHubProperties.java
```diff
@@ -30,6 +30,8 @@
 
     private String checkpointAccessKey;
 
+    private String checkpointContainer;
+
```
jialindai: This seems a confgiuration change, is it backward compatible when user doesn't provide this config?

spring-cloud-azure-dependencies/pom.xml
```diff
@@ -55,7 +55,8 @@
         <hibernate.validator.version>6.1.0.Final</hibernate.validator.version>
         <google.guava.version>20.0</google.guava.version>
         <azure.sdk.version>1.23.0</azure.sdk.version>
-        <azure.eventhubs.eph.version>2.4.0</azure.eventhubs.eph.version>
+        <azure.eventhubs.version>5.0.0-beta.6</azure.eventhubs.version>
+        <azure.eventhubs.checkpoint.store.version>1.0.0-beta.4</azure.eventhubs.checkpoint.store.version>
```
jialindai: When will eventhub sdk release version ready?

PR #637, opened 2020-03-06
Update message usage for 2.0 azure service bus

samukce:
## Description
Update the usage to get message body from Azure Service Bus messages.

## Related PRs
https://github.com/Azure/azure-service-bus-java/pull/296

## Todos
- [x] Tests

## Steps to Test
Creating a new event manually as a String using some tools as Cerebrata Cerulean
image
And, try to consume the message. They will arrive as a VALUE instead of BINARY.

**Other way to simulate is:**
Using this code sample, https://github.com/Azure-Samples/spring-jms-service-bus.git, add  a message converter to instead of use a binary message use a text Message.
```java
@Component
public class UserConverter implements MessageConverter {
    @Override
    public Message toMessage(Object object, Session session)
            throws JMSException {
        String payload = null;
        try {
            payload = new ObjectMapper().writeValueAsString(object);
        } catch (JsonProcessingException e) {
            e.printStackTrace();
        }

        return session.createTextMessage(payload);
    }
...
}
```

### DIFF
--- a/spring-integration-azure/spring-integration-azure-core/src/main/java/com/microsoft/azure/spring/integration/core/converter/AbstractAzureMessageConverter.java
+++ b/spring-integration-azure/spring-integration-azure-core/src/main/java/com/microsoft/azure/spring/integration/core/converter/AbstractAzureMessageConverter.java
@@ -26,7 +26,7 @@ import java.util.Map;
 public abstract class AbstractAzureMessageConverter<T> implements AzureMessageConverter<T> {
     private static ObjectMapper objectMapper = new ObjectMapper();
 
-    private static byte[] toPayload(Object object) {
+    protected static byte[] toPayload(Object object) {
         try {
             return objectMapper.writeValueAsBytes(object);
         } catch (JsonProcessingException e) {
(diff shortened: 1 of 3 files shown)


samukce:
hi @saragluna , could you help to take a look on this one?

saragluna:
Thanks for your contribution. Do we have a reference doc to help explain the difference between `messageBody()` and `body()` methods? What are the main scenarios will this change be used?

samukce:
@saragluna the official documentation is not the explanatory: https://docs.microsoft.com/en-us/java/api/com.microsoft.azure.servicebus.message.getbody?view=azure-java-stable

> Looks like AMQP can have multiple data sections:
> 
> The body consists of one of the following three choices: one or more data sections, one or more amqp-sequence sections, or a single amqp-value section. #Resolved
https://github.com/Azure/azure-service-bus-java/pull/296#issuecomment-432366924

Also, the AMQP said about the three option: Data, Sequence, and value.

saragluna:
@samukce  

Thanks for the detailed explanation and your contribution!

samukce:
Is there some cadence or eta for the next official release build? @saragluna

saragluna:
We're doing our release every six weeks. The next release will be April 6th.

PR #670, opened 2020-04-22
App Config Store Partial Load Fix

mrm9084:
## Description
Fixes a bug that allowed stores using both /application/ and /<app_name>/ to load when one of them fails, but then stopped refresh from triggering on that store.

### DIFF
--- a/spring-cloud-azure-appconfiguration-config/src/main/java/com/microsoft/azure/spring/cloud/config/AppConfigurationPropertySourceLocator.java
+++ b/spring-cloud-azure-appconfiguration-config/src/main/java/com/microsoft/azure/spring/cloud/config/AppConfigurationPropertySourceLocator.java
@@ -137,28 +137,35 @@ public class AppConfigurationPropertySourceLocator implements PropertySourceLoca
         // There is only one Feature Set for all AppConfigurationPropertySources
         FeatureSet featureSet = new FeatureSet();
 
+        List<AppConfigurationPropertySource> sourceList = new ArrayList<AppConfigurationPropertySource>();
+
         // Reverse in order to add Profile specific properties earlier, and last profile
         // comes first
         Collections.reverse(contexts);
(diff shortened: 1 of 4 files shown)


spring-cloud-azure-appconfiguration-config/src/main/java/com/microsoft/azure/spring/cloud/config/AppConfigurationPropertySourceLocator.java
```diff
@@ -137,28 +137,35 @@ private void addPropertySource(CompositePropertySource composite, ConfigStore st
...
                 } else {
-                    LOGGER.warn("Unable to load configuration from Azure AppConfiguration Service for " + sourceContext,
+                    LOGGER.warn(
+                            "Unable to load configuration from Azure AppConfiguration store " + store.getEndpoint()
+                                    + ". The configurations starting with " + sourceContext + "failed to load.",
                             e);
                     StateHolder.setLoadState(store.getEndpoint(), false);
                 }
+                // If anything breaks we skip out on loading the rest of the store.
+                return;
```
saragluna, line 165: If we return here, we'll not be able to load the store.

mrm9084: That is the idea. If a user uses both /application/ and /application_name/. One can load while the other does not. This causes confusion and error.

mrm9084: That is the idea. If a user uses both /application/ and /application_name/. One can load while the other does not. This causes confusion and error.

saragluna:
Please help resolve the conflicts.

PR #684, opened 2020-04-29
App config1.1.4

mrm9084:
## Description
Fixes same issue as https://github.com/microsoft/spring-cloud-azure/pull/683

Not sure why, but this is pulling in all of the other 1.1.4 changes are also listed.

### DIFF
--- a/spring-cloud-azure-appconfiguration-config/src/main/java/com/microsoft/azure/spring/cloud/config/AppConfigurationBootstrapConfiguration.java
+++ b/spring-cloud-azure-appconfiguration-config/src/main/java/com/microsoft/azure/spring/cloud/config/AppConfigurationBootstrapConfiguration.java
@@ -6,13 +6,12 @@
 package com.microsoft.azure.spring.cloud.config;
 
 import java.util.List;
+import java.util.Optional;
 
 import org.apache.http.impl.client.CloseableHttpClient;
 import org.apache.http.impl.client.HttpClients;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.springframework.beans.factory.NoSuchBeanDefinitionException;
-import org.springframework.beans.factory.NoUniqueBeanDefinitionException;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
@@ -67,31 +66,50 @@ public class AppConfigurationBootstrapConfiguration {
 
     @Bean
     public AppConfigurationPropertySourceLocator sourceLocator(AppConfigurationProperties properties,
-            AppConfigurationProviderProperties appProperties, ClientStore clients, ApplicationContext context) {
+            AppConfigurationProviderProperties appProperties, ClientStore clients, ApplicationContext context,
+            Optional<KeyVaultCredentialProvider> keyVaultCredentialProviderOptional,
+            Optional<SecretClientBuilderSetup> keyVaultClientProviderOptional) {
+        
         KeyVaultCredentialProvider keyVaultCredentialProvider = null;
-        try {
-            keyVaultCredentialProvider = context.getBean(KeyVaultCredentialProvider.class);
-        } catch (NoUniqueBeanDefinitionException e) {
-            throw new RuntimeException("Failed to find unique KeyVaultCredentialProvider Bean for authentication.", e);
-        } catch (NoSuchBeanDefinitionException e) {
+        SecretClientBuilderSetup keyVaultClientProvider = null;
+        
+        if (!keyVaultCredentialProviderOptional.isPresent()) {
             LOGGER.debug("No KeyVaultCredentialProvider found.");
+        } else {
+           keyVaultCredentialProvider = keyVaultCredentialProviderOptional.get(); 
         }
+
+        if (!keyVaultClientProviderOptional.isPresent()) {
+            LOGGER.debug("No KeyVaultCredentialProvider found.");
+        } else {
+           keyVaultClientProvider = keyVaultClientProviderOptional.get(); 
+        }
+        
         return new AppConfigurationPropertySourceLocator(properties, appProperties, clients,
-                keyVaultCredentialProvider);
+                keyVaultCredentialProvider, keyVaultClientProvider);
     }
 
     @Bean
     public ClientStore buildClientStores(AppConfigurationProperties properties,
-            AppConfigurationProviderProperties appProperties, ConnectionPool pool, ApplicationContext context) {
+            AppConfigurationProviderProperties appProperties, ConnectionPool pool, ApplicationContext context,
+            Optional<AppConfigurationCredentialProvider> tokenCredentialProviderOptional,
+            Optional<ConfigurationClientBuilderSetup> clientProviderOptional) {
+        
         AppConfigurationCredentialProvider tokenCredentialProvider = null;
-        try {
-            tokenCredentialProvider = context.getBean(AppConfigurationCredentialProvider.class);
-        } catch (NoUniqueBeanDefinitionException e) {
-            throw new RuntimeException(
-                    "Failed to find unique AppConfigurationCredentialProvider Bean for authentication.", e);
-        } catch (NoSuchBeanDefinitionException e) {
+        ConfigurationClientBuilderSetup clientProvider = null;
+        
+        if (!tokenCredentialProviderOptional.isPresent()) {
             LOGGER.debug("No AppConfigurationCredentialProvider found.");
+        } else {
+            tokenCredentialProvider = tokenCredentialProviderOptional.get();
         }
-        return new ClientStore(appProperties, pool, tokenCredentialProvider);
+        
+        if (!clientProviderOptional.isPresent()) {
+            LOGGER.debug("No AppConfigurationClientProvider found.");
+        } else {
+            clientProvider = clientProviderOptional.get();
+        }
+        
+        return new ClientStore(appProperties, pool, tokenCredentialProvider, clientProvider);
     }
 }
--- a/spring-cloud-azure-appconfiguration-config/src/main/java/com/microsoft/azure/spring/cloud/config/AppConfigurationPropertySource.java
+++ b/spring-cloud-azure-appconfiguration-config/src/main/java/com/microsoft/azure/spring/cloud/config/AppConfigurationPropertySource.java
@@ -58,13 +58,16 @@ public class AppConfigurationPropertySource extends EnumerablePropertySource<Con
 
     private KeyVaultCredentialProvider keyVaultCredentialProvider;
 
+    private SecretClientBuilderSetup keyVaultClientProvider;
+
     private AppConfigurationProviderProperties appProperties;
 
     private ConfigStore configStore;
 
     AppConfigurationPropertySource(String context, ConfigStore configStore, String label,
             AppConfigurationProperties appConfigurationProperties, ClientStore clients,
-            AppConfigurationProviderProperties appProperties, KeyVaultCredentialProvider keyVaultCredentialProvider) {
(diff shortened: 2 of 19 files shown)


spring-cloud-azure-appconfiguration-config/src/main/java/com/microsoft/azure/spring/cloud/config/AppConfigurationBootstrapConfiguration.java
```diff
@@ -67,31 +66,50 @@ public CloseableHttpClient closeableHttpClient() {
 
     @Bean
     public AppConfigurationPropertySourceLocator sourceLocator(AppConfigurationProperties properties,
-            AppConfigurationProviderProperties appProperties, ClientStore clients, ApplicationContext context) {
+            AppConfigurationProviderProperties appProperties, ClientStore clients, ApplicationContext context,
+            Optional<KeyVaultCredentialProvider> keyVaultCredentialProviderOptional,
+            Optional<SecretClientBuilderSetup> keyVaultClientProviderOptional) {
```
samukce, line 71: It looks like not good practice to use Optionals as public method arguments. This forces the client used to add extra logics and unnecessary complexity. I like this article that talks about it http://dolszewski.com/java/java-8-optional-use-cases/.
If we want to be protected against possible NPE what do you think if we have the optional use in another way such as

```java 
public AppConfigurationPropertySourceLocator sourceLocator(AppConfigurationProperties properties,
            AppConfigurationProviderProperties appProperties, 
            ClientStore clients, 
            ApplicationContext context,
            KeyVaultCredentialProvider keyVaultCredentialProvider,
            SecretClientBuilderSetup keyVaultClientProvider) {
            
      Optional<KeyVaultCredentialProvider> keyVaultCredentialProviderOptional = offNullable(keyVaultCredentialProvider);
      Optional<SecretClientBuilderSetup> keyVaultClientProviderOptional = offNullable(keyVaultClientProvider);
....
```

spring-cloud-azure-appconfiguration-config/src/main/java/com/microsoft/azure/spring/cloud/config/AppConfigurationBootstrapConfiguration.java
```diff
@@ -67,31 +66,50 @@ public CloseableHttpClient closeableHttpClient() {
...
             LOGGER.debug("No KeyVaultCredentialProvider found.");
+        } else {
+           keyVaultCredentialProvider = keyVaultCredentialProviderOptional.get(); 
         }
+
+        if (!keyVaultClientProviderOptional.isPresent()) {
+            LOGGER.debug("No KeyVaultCredentialProvider found.");
+        } else {
+           keyVaultClientProvider = keyVaultClientProviderOptional.get(); 
+        }
```
samukce, line 86: To read the affirmative statement it's used to be easier than the negative statement, so what do you think about revert this if/else condition? (Sugar syntax but it's helpful for a more natural reading )
```suggestion
        if (keyVaultClientProviderOptional.isPresent()) {
           keyVaultClientProvider = keyVaultClientProviderOptional.get(); 
        } else {
           LOGGER.debug("No KeyVaultCredentialProvider found.");
        }
```

spring-cloud-azure-appconfiguration-config/src/main/java/com/microsoft/azure/spring/cloud/config/AppConfigurationBootstrapConfiguration.java
```diff
@@ -67,31 +66,50 @@ public CloseableHttpClient closeableHttpClient() {
...
+        
+        if (!clientProviderOptional.isPresent()) {
+            LOGGER.debug("No AppConfigurationClientProvider found.");
+        } else {
+            clientProvider = clientProviderOptional.get();
         }
-        return new ClientStore(appProperties, pool, tokenCredentialProvider);
+        
+        return new ClientStore(appProperties, pool, tokenCredentialProvider, clientProvider);
```
samukce, line 113: One idea to avoid an intermediate variable it would be to use the `Optional.else` directly here
```suggestion
return new ClientStore(
                appProperties, 
                pool, 
                tokenCredentialProviderOptional.orElse(null), 
                clientProviderOptional.orElse(null)
        );
```

spring-cloud-azure-appconfiguration-config/src/main/java/com/microsoft/azure/spring/cloud/config/AppConfigurationRefresh.java
```diff
@@ -110,6 +113,11 @@ private boolean refreshStores() {
                 if (willRefresh) {
                     // Only one refresh Event needs to be call to update all of the
                     // stores, not one for each.
+                    if (eventDataInfo.equals("*")) {
+                        LOGGER.info("Configuration Refresh event triggered by store modification.");
+                    } else {
+                        LOGGER.info("Configuration Refresh Event triggered by " + eventDataInfo);
+                    }
```
samukce, line 120: Instead of having an if/else condition for log something adding an extra logic what do you think about simplifying this one? 
```suggestion
                    LOGGER.info("Configuration Refresh event triggered by {}.", eventDataInfo);
```

spring-cloud-azure-appconfiguration-config/src/main/java/com/microsoft/azure/spring/cloud/config/AppConfigurationPropertySource.java
```diff
@@ -108,10 +112,7 @@ public Object getProperty(String name) {
...
         String storeName = configStore.getEndpoint();
         Date date = new Date();
-        SettingSelector settingSelector = new SettingSelector();
-        if (!label.equals("%00")) {
-            settingSelector.setLabelFilter(label);
-        }
+        SettingSelector settingSelector = new SettingSelector().setLabelFilter(label);
 
         // * for wildcard match
         settingSelector.setKeyFilter(context + "*");
```
samukce, line 118: Can we have this symbol `"*"` in some static variable with some meaningful name saying what means it?

spring-cloud-azure-appconfiguration-config/src/main/java/com/microsoft/azure/spring/cloud/config/AppConfigurationProviderProperties.java
```diff
@@ -28,7 +28,7 @@
     private String version;
 
     @NotNull
-    @Value("${maxRetries:12}")
+    @Value("${maxRetries:2}")
```
samukce, line 31: Could you share why such a huge difference from the prior value? From 12 to 2.

spring-cloud-azure-appconfiguration-config/src/main/java/com/microsoft/azure/spring/cloud/config/ConfigurationClientBuilderSetup.java
```diff
@@ -0,0 +1,15 @@
...
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See LICENSE in the project root for
+ * license information.
+ */
+package com.microsoft.azure.spring.cloud.config;
+
+
+import com.azure.data.appconfiguration.ConfigurationClientBuilder;
+
+public interface ConfigurationClientBuilderSetup {
```
samukce, line 11: If we following the name convention from class should finish with substantive
```suggestion
public interface ConfigurationClientSetupBuilder {
```

spring-cloud-azure-appconfiguration-config/src/main/java/com/microsoft/azure/spring/cloud/config/SecretClientBuilderSetup.java
```diff
@@ -0,0 +1,14 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See LICENSE in the project root for
+ * license information.
+ */
+package com.microsoft.azure.spring.cloud.config;
+
+import com.azure.security.keyvault.secrets.SecretClientBuilder;
+
+public interface SecretClientBuilderSetup {
```
samukce, line 10: The same here 
```suggestion
public interface SecretClientSetupBuilder {
```

spring-cloud-azure-appconfiguration-config/src/main/java/com/microsoft/azure/spring/cloud/config/stores/ClientStore.java
```diff
@@ -72,38 +81,51 @@ private ConfigurationAsyncClient buildClient(String store) throws IllegalArgumen
...
      * @param settingSelector Information on which setting to pull. i.e. number of
      * results, key value...
      * @param storeName Name of the App Configuration store to query against.
      * @return List of Configuration Settings.
      */
-    public final List<ConfigurationSetting> listSettingRevisons(SettingSelector settingSelector, String storeName) {
+    public final ConfigurationSetting getRevison(SettingSelector settingSelector, String storeName) {
         ConfigurationAsyncClient client = buildClient(storeName);
-        return client.listRevisions(settingSelector).collectList().block();
+        return client.listRevisions(settingSelector).blockFirst();
```
samukce, line 128: Sorry for ask, but do we have how to know if the first in this list is the latest configuration from the revisions?

spring-cloud-azure-appconfiguration-config/src/main/java/com/microsoft/azure/spring/cloud/config/stores/ClientStore.java
```diff
@@ -139,7 +161,7 @@ public String watchedKeyNames(ConfigStore store, Map<String, List<String>> store
         String watchedKey = store.getWatchedKey().trim();
         List<String> contexts = storeContextsMap.get(store.getEndpoint());
 
-        String watchedKeys = contexts.stream().map(ctx -> genKey(ctx, watchedKey))
+        String watchedKeys = contexts.stream().map(ctx -> genKey(ctx, watchedKey)).distinct()
                 .collect(Collectors.joining(","));
 
         if (watchedKeys.contains(",") && watchedKeys.contains("*")) {
```
samukce, line 167: If seems this logic could be simplified. If I understood right, the logic here is: 
```
If we have more than one(not equal) or `!StringUtils.isNoneEmpty(watchedKey)`
     return "*"
else 
    return context+watchedKey
```

What do you think about this code to replace the logic from line 164 to 180?
```java
    private static final String ALL_KEYS = "*";
    ...
    private String supportedWatchedKey(List<String> contexts, String watchedKey) {
        if (watchedKey.contains(ALL_KEYS)){
            return ALL_KEYS;
        }

        List<String> contextsUniques = contexts.stream().distinct().collect(toList());
        if (contextsUniques.size() > 1) {
            LOGGER.warn("Multi keys not supported. It will use watch all keys(*) instead {}", Arrays.toString(contextsUniques.toArray()));
            return ALL_KEYS;
        }

        return contextsUniques.stream().findFirst().orElse("") + watchedKey;
    }
```

spring-cloud-azure-appconfiguration-config/src/test/java/com/microsoft/azure/spring/cloud/config/stores/ConfigStoreTest.java
```diff
@@ -26,5 +27,21 @@ public void invalidEndpoint() {
...
+        configStore.setLabel("dev");
+        assertEquals(configStore.getLabels()[0], "dev");
+        
+        configStore.setLabel("dev,test");
+        assertEquals(configStore.getLabels()[0], "test");
+        assertEquals(configStore.getLabels()[1], "dev");
+        
+        configStore.setLabel(",");
+        assertEquals(configStore.getLabels()[0], "\0");
+    }
```
samukce, line 45: Can we breakdown this test in others 4 saying what the expected behavior for each one? This way, if some tests break we can see easily which behavior it was impacted for some change in the future.

For instance: `shouldStartLabelAsZeroCharacter`,  `shouldReplaceLabelBySingleValue`, `shouldSplitLabelByComma`....

Issue #698, opened 2020-05-13
small typo in the docs, property name is sessionsEnabled


leassis:
This is very simple but leads to a few minutes of a resource. In documentation says that the property is named as sessionEnabled but the real name is sessionsEnabled (session in plural)

yiliuTo:
Thanks for your reporting. Would you like to share a link of the documentation that uses the _sessionEnabled_ property?

leassis:
Sorry for late reply, this is the page that I've mentioned

https://github.com/microsoft/spring-cloud-azure/tree/master/spring-cloud-azure-stream-binder/spring-cloud-azure-servicebus-topic-stream-binder

yiliuTo:
Thank you, we will deal with this issue~

saragluna:
Closing this issue.

PR #708, opened 2020-05-28
modify readme for topic and queue stream binders for typos

yiliuTo:
update typos that in the readme.md of topic and queue stream binder, property name should be _sessionsEnabled_ instead of _sessionEnabled_.


saragluna:
#698 fixed by this PR.

Issue #675, opened 2020-04-25
With multiple labels and single watched-key configured for app config, app configuration throws error while application start up


Karthickramk:
## Expected Behavior
* Application should start fine with the below configuration.
```
spring:
  cloud:
    azure:
      appconfiguration:
        default-context: Sams-Cart
        cache-expiration: 60000
        stores:
          - connection-string: ${appConfig-scus.${spring.profiles.active:dev}}
            label: default, release1
            watched-key: appconfig-refresh-key
          - connection-string: ${appConfig-wus.${spring.profiles.active:dev}}
            label: default

```
## Current Behavior
* But while starting the application, app failed with the below error.

```
25/Apr/2020 18:03:32:908 [main]  INFO  com.azure.data.appconfiguration.ConfigurationAsyncClient:204 - Listing ConfigurationSetting revisions - SettingSelector(keyFilter=/Sams-Cart_local/appconfig-refresh-key,/Sams-Cart_local/appconfig-refresh-key,/Sams-Cart/appconfig-refresh-key,/Sams-Cart/appconfig-refresh-key,/Sams-Cart_local/appconfig-refresh-key,/Sams-Cart_local/appconfig-refresh-key, labelFilter=*, acceptDateTime=null, fields=ALL_FIELDS)
25/Apr/2020 18:03:33:189 [reactor-http-nio-2]  WARN  com.azure.data.appconfiguration.ConfigurationAsyncClient:210 - Failed to list ConfigurationSetting revisions - SettingSelector(keyFilter=/Sams-Cart_local/appconfig-refresh-key,/Sams-Cart_local/appconfig-refresh-key,/Sams-Cart/appconfig-refresh-key,/Sams-Cart/appconfig-refresh-key,/Sams-Cart_local/appconfig-refresh-key,/Sams-Cart_local/appconfig-refresh-key, labelFilter=*, acceptDateTime=null, fields=ALL_FIELDS)
Status code 400, "{"type":"https://azconfig.io/errors/invalid-argument","title":"Invalid request parameter 'key'","name":"key","detail":"key(0): Too many values. Maximum supported is 5","pos":0,"status":400}"
25/Apr/2020 18:03:33:193 [main]  ERROR com.microsoft.azure.spring.cloud.config.AppConfigurationPropertySourceLocator:151 - Fail fast is set and there was an error reading configuration from Azure App Configuration Service for /Sams-Cart_local/
[...]

PR #723, opened 2020-07-06
App Configuration Bug Fixes

mrm9084:
- Fixes Refresh not being disabled when provider is disabled. https://github.com/Azure/AppConfiguration/issues/332
- Fixed an issue where duplicate watch keys were added to watch checks, the number of duplicates were based off the number labels and profiles used. This might fix https://github.com/microsoft/spring-cloud-azure/issues/675
- Fixed feature management name issue.
- Fixed issue where switching from an on/off feature flag to conditional didn't work.


jialindai:
Hi Matt, the target branch should be appconfiguration/1.2.x, so future changes on 1.2.x can apply here.

PR #734, opened 2020-07-15
remove storage entry in autoconfigure module

yiliuTo:
remove storage autoconfiguration entry in autofigure spring.factories


spring-cloud-azure-starters/spring-cloud-starter-azure-eventhubs-kafka/pom.xml
```diff
@@ -23,6 +23,10 @@
             <groupId>org.springframework.cloud</groupId>
             <artifactId>spring-cloud-starter-stream-kafka</artifactId>
         </dependency>
+        <dependency>
```
saragluna, line 26: Seems like we don't need this If we remove the auto-configuration entry for the context in spring.factories?

PR #759, opened 2020-11-09
fix broken links to azure-spring-cloud-stream-binder projects

ulrichwinter:
The Readme at https://github.com/ulrichwinter/spring-cloud-azure contains three links to stream binder projects, which have apparently been changed since the Readme was authored.


saragluna (approved):
LGTM. Thanks for helping change these!

PR #793, opened 2021-12-14
Fix EventHub binding consumer property name

avpines:
Fix binding consumer property name (track-last-enqueued-event-properties)


chenrujun (approved):
LGTM.

PR #797, opened 2021-12-20
Add description about managed identity types 

saragluna:
As title.

### DIFF
--- a/docs/src/main/asciidoc/authentication.adoc
+++ b/docs/src/main/asciidoc/authentication.adoc
@@ -12,6 +12,7 @@
 :azure-storage-queue-data-reader: https://docs.microsoft.com/azure/role-based-access-control/built-in-roles#storage-queue-data-reader
 :key-vault-access-policy: https://docs.microsoft.com/azure/key-vault/general/assign-access-policy
 :cosmos-db-rbac: https://docs.microsoft.com/azure/cosmos-db/how-to-setup-rbac
+:managed-identity: https://docs.microsoft.com/azure/active-directory/managed-identities-azure-resources/overview
 [#authentication]
 == Authentication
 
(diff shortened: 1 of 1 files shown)


docs/src/main/asciidoc/authentication.adoc
```diff
@@ -42,6 +42,14 @@ A common challenge for developers is the management of secrets and credentials u
...
 
+==== Managed Identity Types
+There are two types of managed identities:
+
+- System-assigned Some Azure services allow you to enable a managed identity directly on a service instance. When you enable a system-assigned managed identity an identity is created in Azure AD that is tied to the lifecycle of that service instance. So when the resource is deleted, Azure automatically deletes the identity for you. By design, only that Azure resource can use this identity to request tokens from Azure AD.
+- User-assigned You may also create a managed identity as a standalone Azure resource. You can create a user-assigned managed identity and assign it to one or more instances of an Azure service. In the case of user-assigned managed identities, the identity is managed separately from the resources that use it.
```
chenrujun, line 49: Suggestion:

`- System-assigned` -> `- **System-assigned**`
`- User-assigned` -> `- **User-assigned**`

image

chenrujun: Adding a link is better than copy text from MS docs. Because MS docs may update in the future. We should keep single source of truth.

MS docs: https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/overview#managed-identity-types

docs/src/main/asciidoc/authentication.adoc
```diff
@@ -42,6 +42,14 @@ A common challenge for developers is the management of secrets and credentials u
...
+
+NOTE: When a user-assigned managed identity is used, you can specify it by `spring.cloud.azure.credential.managed-identity-client-id` or `spring.cloud.azure.<azure-service>.credential.managed-identity-client-id`.
```
chenrujun, line 51: Do we need this: `When system-assigned managed identity is used, ...`

chenrujun (changes requested):

docs/src/main/asciidoc/authentication.adoc
saragluna: yes, if users is looking for configuration properties related to MI, this will help them.

docs/src/main/asciidoc/authentication.adoc
saragluna: I prefer to explain these concepts with minimal description, but I agree that we can link to MS Docs. But what came first to me is that jumping out of the current doc is annoying.

docs/src/main/asciidoc/authentication.adoc
saragluna: But no extra configuration is needed when system-assigned mi is used.

docs/src/main/asciidoc/authentication.adoc
chenrujun: Could you please add that link?

docs/src/main/asciidoc/authentication.adoc
chenrujun: Could you please add content about that? Let reader known that.

chenrujun (approved):
LGTM.

PR #799, opened 2021-12-24
Add content about version mapping.

chenrujun:
As title.

Render result:
image

And version is not displayed in above image, but it will be rendered, too.
Just like current page: 
https://microsoft.github.io/spring-cloud-azure/docs/current/reference/html/index.html
image


chenrujun:
Hi, @stliu , @saragluna , please help to review this page when you have time.

saragluna (approved):
LGTM

PR #803, opened 2022-01-13
Make sure all contents in readme is included for secret management module

chenrujun:
As title.

### DIFF
--- a/docs/src/main/asciidoc/legal.adoc
+++ b/docs/src/main/asciidoc/legal.adoc
@@ -1,1 +1,4 @@
+
+(C) 2016-2021 the original authors.
+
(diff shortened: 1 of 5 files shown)


docs/src/main/asciidoc/legal.adoc
```diff
@@ -1 +1,4 @@
+
+(C) 2016-2021 the original authors.
```
backwind1233, line 2: 2016-2021 --> 2016-2022?

backwind1233: what's this ?

backwind1233 (approved):
LGTM

docs/src/main/asciidoc/legal.adoc
chenrujun: https://docs.spring.io/spring-framework/docs/current/reference/html/

image

PR #817, opened 2022-01-20
Fix/version not right

chenrujun:
1. Use variable instead of plain text about version in adoc files. 
2. Delete unnecessary "<version>" in pom. Extend from parent instead.
3. Change markdown style title to adoc style.
4. Fix error content about Azure AD related links.
5. Fix the problem: Configuration properties not displayed in appendix.

### DIFF
--- a/docs/pom.xml
+++ b/docs/pom.xml
@@ -13,7 +13,6 @@
 	<name>spring-cloud-azure-build-docs</name>
 	<packaging>jar</packaging>
 	<description>Spring Cloud Azure Build Docs</description>
-	<version>4.0.0-beta.3</version>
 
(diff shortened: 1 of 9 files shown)


docs/pom.xml
```diff
@@ -13,7 +13,6 @@
 	<name>spring-cloud-azure-build-docs</name>
 	<packaging>jar</packaging>
 	<description>Spring Cloud Azure Build Docs</description>
-	<version>4.0.0-beta.3</version>
```
chenrujun, line 16: "<version>" can inherit from parent.

saragluna (approved):
LGTM.

PR #828, opened 2022-01-21
Docs App Configuration - Key Values & Health Indicator

mrm9084:
Adds doc sections for Key Values: Placeholders, Json, Key Vault Reference and Health Indicator.

### DIFF
--- a/docs/src/main/asciidoc/health-indicator.adoc
+++ b/docs/src/main/asciidoc/health-indicator.adoc
@@ -1,0 +1,9 @@
+## Health Indicator
+
+The client library comes with a Health Indicator that check whether the connection to the Azure App Configuration store(s) is healthy or not. If enabled, for each store it gives a status of:
+
+ * UP - The last connection was successful
+ * DOWN- The last connection resulted in a non 200 error code
+ * NOT LOADED - The config store is listed in the local configuration file, but wasn't loaded from at startup. This can either be because it is disabled, or failed to load at startup and the `fail-fast` configuration for the store was set to false.
+
+You can enable the Health Indicator by setting `management.health.azure-app-configuration.enabled=true`.
--- a/docs/src/main/asciidoc/key-values.adoc
+++ b/docs/src/main/asciidoc/key-values.adoc
@@ -1,4 +1,45 @@
-== Key Vault References
+== Key Values
+
+Azure App Configuration Supports multiple types of key values, some of which have special features built into them.
+
+=== Placeholders
+
+The client library supports configurations with `${}`-style environment placeholders. If referencing a Azure App Configuration key with a placeholder any prefix needs to be removed in the reference. For example: `/application/config.message` is referenced as `${config.message}`.
+
+NOTE: The prefix being removed matches the value `spring.cloud.azure.appconfiguration.stores[0].selects[0].key-filter`.
+
+=== JSON
+
+Configurations that have the a content-type `application/json` are processed as json objects. This enables one configuration mapping to a complex object in a `@Configuration`. The json key `/application/config.colors` with the value
+
+[source,json,indent=0]
+----
+{
+	"Red": {
+		"value": [255, 0, 0]
+	},
+	"Blue": {
+		"value": [0, 255, 0]
+	},
+	"Green": {
+		"value": [0, 0, 255]
+	}
+}
+----
+
+maps to
+
+[source,java,indent=0]
+----
+@ConfigurationProperties(prefix = "config")
+public class MyConfigurations {
+
+    private HashMap<String, Color> colors;
+
+}
+----
+
+=== Key Vault References
 
 Azure App Configuration and its client libraries support referencing secrets stored in Key Vault. In App Configuration keys can be created the have values that are URIs that map to secrets in a Key Vault. This enables using secrets as configuration without storing them as configurations.
 
@@ -6,7 +47,7 @@ Your application uses the client provider to retrieve Key Vault references, just
 
 NOTE: Key Vault only allows for secrets to be retrieved one at a time, so each key vault reference stored in App Configuration will result in a pull against Key Vault.
 
-=== Creating Key Vault References
+==== Creating Key Vault References
 
 You can easily create a Key Vault reference in the Azure Portal in the Configuration explorer using the Create -> Key Vault reference option. You will be able to select a secret to make a reference to from any Key Vault you have access to. You can also create arbitrary Key Vault references using the Input option.
 
@@ -17,7 +58,7 @@ You can also create a Key Vault reference through the cli using:
 az appconfig kv set-keyvault --name <name-of-your-store> --key <key-name> --secret-identifier <uri-to-your-secret>
 ----
 
-=== Using Key Vault References
+==== Using Key Vault References
 
 If you are using User Assigned or System Assigned Identity all you need to do is make sure that identity also has access to the Key Vault and reading secrets.
 
@@ -41,7 +82,7 @@ public class MyCredentials implements KeyVaultCredentialProvider {
 
 Using the KeyVaultCredentialProvider you can provide any TokenCredential type supported by the link:https://github.com/Azure/azure-sdk-for-java/tree/main/sdk/identity/azure-identity#credential-classes[Azure Identity Library].
 
-=== Resolve Non Key Vault Secrets
+==== Resolve Non Key Vault Secrets
 
 The App Configuration client provides a method of locally resolving secrets that don't have a Key Vault associated with them. This is done through the `KeyVaultSecretProvider`.
 
--- a/docs/src/main/asciidoc/getting-help.adoc
+++ b/docs/src/main/asciidoc/getting-help.adoc
@@ -8,10 +8,10 @@ If you have any questions about this doc, please ask by creating GitHub issue. A
 |===
 |GitHub repositories | Description
 
-|https://github.com/Azure/azure-sdk-for-java/tree/main/sdk/spring[Azure/azure-sdk-for-java
+|https://github.com/Azure/azure-sdk-for-java/tree/main/sdk/spring[Azure/azure-sdk-for-java]
 |This repository used to hold the source code.
 
-|https://github.com/Azure/AppConfiguration]
+|https://github.com/Azure/AppConfiguration
 |This repository used to hold release notes, and general Azure App Configuration Issues.
 
 |https://github.com/microsoft/spring-cloud-azure/tree/docs/docs/src/main/asciidoc[microsoft/spring-cloud-azure]
--- a/docs/src/main/asciidoc/index.adoc
+++ b/docs/src/main/asciidoc/index.adoc
@@ -18,4 +18,6 @@ include::loading-configurations.adoc[]
 
 include::authentication.adoc[]
 
+include::key-vault-references.adoc[]
+
 include::forMoreInfo.adoc[]
--- a/docs/src/main/asciidoc/index.adoc
+++ b/docs/src/main/asciidoc/index.adoc
@@ -18,6 +18,8 @@ include::loading-configurations.adoc[]
 
 include::authentication.adoc[]
 
-include::key-vault-references.adoc[]
+include::key-values.adoc[]
+
+include::health-indicator.adoc[]
 
 include::forMoreInfo.adoc[]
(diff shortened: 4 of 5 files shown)


docs/src/main/asciidoc/health-indicator.adoc
```diff
@@ -0,0 +1,9 @@
+## Health Indicator
+
+The client library comes with a Health Indicator that check whether the connection to the Azure App Configuration store(s) is healthy or not. If enabled, for each store it gives a status of:
```
AlexandraKemperMS, line 3: Nits: check -> checks 
comma after store instead of enabled

docs/src/main/asciidoc/health-indicator.adoc
```diff
@@ -0,0 +1,9 @@
...
+
+ * UP - The last connection was successful
```
AlexandraKemperMS, line 5: could you also state what actions can be taken by a customer if they see each of these? for example, for the down connection, you should follow the steps found at (link) or send another request etc.

docs/src/main/asciidoc/health-indicator.adoc
```diff
@@ -0,0 +1,9 @@
...
+ * DOWN- The last connection resulted in a non 200 error code
+ * NOT LOADED - The config store is listed in the local configuration file, but wasn't loaded from at startup. This can either be because it is disabled, or failed to load at startup and the `fail-fast` configuration for the store was set to false.
```
AlexandraKemperMS, line 7: do you mean "from the file"? 
this is not clear. are the reasons this can happen because 
1. it is disabled 
2. failed to load at startup and the `fail-fast` configuration for the store was set to false
or because 
1. it is disabled 
2. failed to load at startup
 and in both situations the `fail-fast` configuration for the store was set to false

saragluna (approved):
LGTM, please resolve conversations before merging.

docs/src/main/asciidoc/health-indicator.adoc
mrm9084: I am not sure which steps we would suggest for down as the client library would be retried. Or it may be a service issue. Do we have other docs for this?

docs/src/main/asciidoc/health-indicator.adoc
mrm9084: 1 & 2, any ideas how to clear this up?

chenrujun:
@mrm9084 , could you please merge this PR?

mrm9084:
> @mrm9084 , could you please merge this PR?

@AlexandraKemperMS has only been able to look at one of the files. The other file still needs to be reviewed.

docs/src/main/asciidoc/health-indicator.adoc
AlexandraKemperMS: I don't know any specific docs for this, but we could at least append the sentence and say something like ".., this could be due to a number of issues, such as the client library being retired or there being a service issue"

docs/src/main/asciidoc/health-indicator.adoc
AlexandraKemperMS: Maybe this: 
The config store is listed in the local configuration file, but the config wasn't loaded from the file at startup. This can be because of two reasons. Either loading is disabled, which can be fixed by ___x__y__z_. Otherwise, config failed to load at startup while the `fail-fast` configuration for the store was set to false. This can be fixed by ______setting failfast to true using x command____.

docs/src/main/asciidoc/key-values.adoc
```diff
@@ -0,0 +1,101 @@
+== Key Values
+
+Azure App Configuration Supports multiple types of key values, some of which have special features built into them.
+
+=== Placeholders
+
+The client library supports configurations with `${}`-style environment placeholders. If referencing a Azure App Configuration key with a placeholder any prefix needs to be removed in the reference. For example: `/application/config.message` is referenced as `${config.message}`.
```
AlexandraKemperMS, line 7: What are these configurations/environment variables used for? < answer this in a sentence after the first to make this clear. Ex: "These placeholders can be used to separate between environments such as prod, dev, and test, and let you easily switch between them"

docs/src/main/asciidoc/key-values.adoc
```diff
@@ -0,0 +1,101 @@
...
+public class MyConfigurations {
+
+    private HashMap<String, Color> colors;
+
+}
+----
+
+=== Key Vault References
+
+Azure App Configuration and its client libraries support referencing secrets stored in Key Vault. In App Configuration keys can be created the have values that are URIs that map to secrets in a Key Vault. This enables using secrets as configuration without storing them as configurations.
```
AlexandraKemperMS, line 44: Second sentence is not clear. What about this: 
" In App Configuration, keys can be created which have URI values that map to a secret stored in a Key Vault. The secrets are securely stored in Key Vault, but can be accessed with the same syntax as configuration"

docs/src/main/asciidoc/key-values.adoc
```diff
@@ -0,0 +1,101 @@
...
+
+Your application uses the client provider to retrieve Key Vault references, just as it does for any other keys stored in App Configuration. Because the client recognizes the keys as Key Vault references, they have a unique content-type, the client will connect to Key Vault to retrieve their values.
```
AlexandraKemperMS, line 46: Nit: "content-type, and the client ..."
"...retrieve their values for you"

docs/src/main/asciidoc/key-values.adoc
```diff
@@ -0,0 +1,101 @@
...
+
+NOTE: Key Vault only allows for secrets to be retrieved one at a time, so each key vault reference stored in App Configuration will result in a pull against Key Vault.
```
AlexandraKemperMS, line 48: Is Key Vault upper case or lower case? Needs to be consistent

docs/src/main/asciidoc/key-values.adoc
```diff
@@ -0,0 +1,101 @@
...
+
+==== Creating Key Vault References
+
+You can easily create a Key Vault reference in the Azure Portal in the Configuration explorer using the Create -> Key Vault reference option. You will be able to select a secret to make a reference to from any Key Vault you have access to. You can also create arbitrary Key Vault references using the Input option.
```
AlexandraKemperMS, line 52: For clarity: "...  select a secret to reference, this can be from any of Key Vaults you have access to>"

what is an arbitrary reference?

docs/src/main/asciidoc/key-values.adoc
```diff
@@ -0,0 +1,101 @@
...
+
+You can also create a Key Vault reference through the cli using:
```
AlexandraKemperMS, line 54: is there a tab system you could use here instead? People will usually try only one of these ways

docs/src/main/asciidoc/key-values.adoc
```diff
@@ -0,0 +1,101 @@
...
+
+[source,azurecli,indent=0]
+----
+az appconfig kv set-keyvault --name <name-of-your-store> --key <key-name> --secret-identifier <uri-to-your-secret>
+----
+
+==== Using Key Vault References
+
+If you are using User Assigned or System Assigned Identity all you need to do is make sure that identity also has access to the Key Vault and reading secrets.
```
AlexandraKemperMS, line 63: Nit: " ... a User Assigned or ..." 
 " to the Key Vault and is able to read secrets". What can customers do if it can't? You should suggest to them that they need to revisit their permissions

docs/src/main/asciidoc/key-values.adoc
```diff
@@ -0,0 +1,101 @@
...
+}
+----
+
+Using the KeyVaultCredentialProvider you can provide any TokenCredential type supported by the link:https://github.com/Azure/azure-sdk-for-java/tree/main/sdk/identity/azure-identity#credential-classes[Azure Identity Library].
+
+==== Resolve Non Key Vault Secrets
+
+The App Configuration client provides a method of locally resolving secrets that don't have a Key Vault associated with them. This is done through the `KeyVaultSecretProvider`.
+
+WARNING: Creating a `KeyVaultSecretProvider` overrides the automatic use of System Assigned Identity. In order to use both `KeyVaultCredentialProvider` needs to be used and return `null` for URI's that need to resolve using `KeyVaultSecretProvider`.
```
AlexandraKemperMS, line 89: Nit: comma after both

docs/src/main/asciidoc/key-values.adoc
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+}
+----
+
+Using the KeyVaultCredentialProvider you can provide any TokenCredential type supported by the link:https://github.com/Azure/azure-sdk-for-java/tree/main/sdk/identity/azure-identity#credential-classes[Azure Identity Library].
+
+==== Resolve Non Key Vault Secrets
+
+The App Configuration client provides a method of locally resolving secrets that don't have a Key Vault associated with them. This is done through the `KeyVaultSecretProvider`.
```
AlexandraKemperMS, line 87: Can we tie this to the code below?

docs/src/main/asciidoc/key-values.adoc
```diff
@@ -0,0 +1,101 @@
+== Key Values
+
+Azure App Configuration Supports multiple types of key values, some of which have special features built into them.
```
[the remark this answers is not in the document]
mrm9084: The multiple types I reference each have a section visible in the main doc, see side bar. https://microsoft.github.io/spring-cloud-azure/docs/azure-app-configuration/2.3.0-beta.1/reference/html/index.html#key-values. So I didn't think we would need one. Thoughts?

docs/src/main/asciidoc/key-values.adoc
mrm9084: That gets a bit award once we merge with the main docs as the go over that in a large section of its own.

docs/src/main/asciidoc/health-indicator.adoc
mrm9084: @AlexandraKemperMS NOT LOADED isn't a failure condition.

When we say loading is disabled it means someone went into the config file and set enabled to false, there is nothing to fix.

fail-fast is mainly meant for testing or having a primary and backup store at startup. If the primary store fails the secondary store is used, but because of this the primary store is never used even on refresh.

The point being there is nothing to "fix". This is more informational.

docs/src/main/asciidoc/key-values.adoc
mrm9084: I think you misunderstand the use of the word environment here. It is environment as environment variable. The wording here is consistent with Springs own documentation on the subject. This isn't a feature we added, we just notifying customers it works exactly the same as in Spring Config and how to use it with our configurations.

docs/src/main/asciidoc/key-values.adoc
mrm9084: The input option allows for any value to be set. It doesn't need to be a key vault, for secret provider.

docs/src/main/asciidoc/key-values.adoc
mrm9084: I can't find a way to get tabs to work with line wrapping. So, I will keep this as is for now.

docs/src/main/asciidoc/health-indicator.adoc
AlexandraKemperMS: Ah, I see, then you can remove the parts around fixing it I mentioned above, but keep the two reasons.

docs/src/main/asciidoc/key-values.adoc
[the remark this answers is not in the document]
AlexandraKemperMS: Yeah, then I think you can just add something to say you are talking about JSON content type, placeholders, and key vault references. but you don't need to mention talking about key vault references in general

docs/src/main/asciidoc/key-values.adoc
AlexandraKemperMS: Could you add this information with maybe a link to where spring defines environmental variables?

docs/src/main/asciidoc/key-values.adoc
AlexandraKemperMS: Makes sense, could you add this to the sentence?

docs/src/main/asciidoc/key-values.adoc
AlexandraKemperMS: Ah ok, then in the future we can link the main docs to this.

docs/src/main/asciidoc/key-values.adoc
mrm9084: These are System Environment variables. Nothing to do with Spring besides that Spring uses them too.

AlexandraKemperMS (approved):

PR #830, opened 2022-01-25
Update eh binder reference doc

yiliuTo:
Update reference doc of spring cloud stream eventhubs:
1. add property of load-balancing
2. refactor configuration section
3. add warning of max pool size breaking change
4. ~add warning of batch consumer content type~ Done before
5. modify the consumer format to accept a message of list for the batch consumer mode
6. add section for message header.


saragluna (approved):
LGTM.

PR #837, opened 2022-02-07
Update sb binder beta2

yiliuTo:
sync #835


chenrujun:
Why merge to `docs` branch?
IMU, we should merge `4.0.0-beta.3` to `4.0.0-beta.4`:
https://github.com/microsoft/spring-cloud-azure/compare/4.0.0-beta.4...4.0.0-beta.3

After `4.0.0-beta.3` released, we do not need to maintain `docs` branch anymore.
Multiple branch support started after 4.0 GA.
Currently, we maintain both `4.0.0-beta.3` and `4.0.0-beta.4` is not because we support multiple version, it's because there is some difference between released version and developing version.

saragluna:
FedEx is using beta.2, this is why docs should be updated for the binders.

saragluna (approved):
LGTM.

PR #838, opened 2022-02-07
update spring integration eventhubs ref doc

yiliuTo:
Update spring integration event hubs ref doc for full feature introduction and configuration options.


yiliuTo:
Close this pr for we only plan to update the latest version

PR #848, opened 2022-02-18
udpate initial position change

yiliuTo:
Add introduction of the breaking change of initial position configuration in eventhubs binder migration guide.
Add example and introduction to initial position configuration in eventhubs binder ref doc.


saragluna (changes requested):

saragluna (approved):
LGTM.

PR #865, opened 2022-03-12
App Configuration - Feature Mangement

mrm9084:
Moved Feature Management from Readme and did a bit of a re-write.

### DIFF
--- a/docs/src/main/asciidoc/feature-management.adoc
+++ b/docs/src/main/asciidoc/feature-management.adoc
@@ -1,0 +1,276 @@
+== Feature Management
+
+Feature Management provide a way for Spring Boot applications dynamically access content. Feature Management has a verity of functions, such as Feature Flags that can enable or disable content, Feature Flags for targeting when content is shown, and Feature Gates for dynamically enabling endpoints.
+
+By default, loading of Feature Flags from Azure App Configuration is disabled. And is enabled through configuration:
+
+[source,properties,indent=0]
+----
+spring.cloud.azure.appconfiguration.stores[0].feature-flags.enabled= true
+----
+
+With Feature Flags enabled they will be loaded into the Spring Configuration system with the prefix `feature-management`. Feature Flags can also be registered in the local configuration file, see <<Feature Flag Declaration>>. The easiest was to use Feature Management is by using the `azure-spring-cloud-feature-management` and `azure-spring-cloud-feature-management-web` libraries. The difference between the two libraries is that `azure-spring-cloud-feature-management-web` takes a dependency on the `spring-web` and `spring-webmvc` libraries to add additional features, such as Feature Gates.
+
+=== Feature Management Basics
+
+==== Feature Flags
+
+Feature flags are composed of two parts, a name and a list of feature-filters that are used to turn the feature on. Feature Flags can either have a boolean state of on/off, or they can have a list of Feature Filters. Feature Flags evaluate Feature Filters until one returns true, if no Feature Filter returns true, then the Feature Flag returns false.
+
+==== Feature Filters
+
+Feature filters define a scenario for when a feature should be enabled. Feature Filters are evaluated synchronously.
+
+As an example, a Microsoft Edge browser feature filter could be designed. This feature filter would activate any features it is attached to as long as an HTTP request is coming from Microsoft Edge.
+
+==== Feature Flag Declaration
+
+The Feature Management library supports Azure App Configuration along with application.yml or bootstrap.yml as sources feature flags. Below is an example of the format used to set up feature flags in a application.yml file.
+
+[source,yaml,indent=0]
+----
+feature-management:
+  feature-t: false
+  feature-u:
+    enabled-for:
+      -
+        name: Random
+  feature-v:
+    enabled-for:
+      -
+        name: TimeWindowFilter
+        parameters:
+          time-window-filter-setting-start: "Wed, 01 May 2019 13:59:59 GMT"
+          time-window-filter-setting-end: "Mon, 01 July 2019 00:00:00 GMT"
+  feature-w:
+    evaluate: false
+    enabled-for:
+      -
+        name: AlwaysOnFilter
+----
+
+`feature-t` is set to false, Feature Flags defined this way just return there value. `feature-u` uses Feature Filters which are defined under `enabled-for`. `feature-u` has one feature filter `Random` which does not require any configuration so it only has the name property. `feature-v` it specifies a feature filter named `TimeWindow`. Feature Filters can be passed parameters to configure the filter. In this case, the start and end times for the feature to be active are configured.
+
+The `AlwaysOnFilter` is a Filter that always evaluates as `true`. The `evaluate` field is used to stop the evaluation of the feature filters, and results in the feature flag to always return `false`.
+
+=== Evaluating Feature Flags
+
+`azure-spring-cloud-feature-management` provides `FeatureManager` to check if a Feature Flag is enabled. It provides an asynchronous way to check the state of the flag.
+
+`azure-spring-cloud-feature-management-web` along with providing `FeatureManager` has `FeatureManagerSnapshot` which caches the state of previously evaluated Feature Flags in the `@RequestScope` to guarantee all requests return the same value. In addition the web library provides `@FeatureGate` which can either block or redirect web requests to different endpoints.
+
+==== Feature Flag Check
+
+`FeatureManager` is a `@Bean` that can be `@Autowired` or injected into `@Component` type objects. `FeatureManager` has a method `isEnabledAsync` that when passed the name of a Feature Flag returns its state.
+
+[source,java,indent=0]
+----
+@Autowired
+FeatureManager featureManager;
+
+if(featureManager.isEnabledAsync("feature-t").block()) {
+    // Do Something
+}
+----
+
+If Feature Management was never configured `isEnabledAsync` or the Feature Flag doesn't exist it will always return false. If an existing Feature Flags is configured with an unknown Feature Filter than an `FilterNotFoundException` is thrown. This can be changed to return false by configuring `fail-fast` to false.
+
+[%autowidth.stretch]
+[options="header"]
+|=== 
+|Name | Description | Required | Default
+|spring.cloud.azure.feature.management.fail-fast | Whether throw RuntimeException or not when exception occurs | No |  true
+|===
+
+`FeatureManagerSnapshot` works the same as `FeatureManager`, besides its caching of results in the `@RequestScope`.
+
+==== Feature Gate
+
+With the Feature Management Web library you can require that a given feature is enabled in order to execute an endpoint. This can be done by using the `@FeatureGate` annotation.
+
+[source,java,indent=0]
+----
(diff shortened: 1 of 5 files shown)


mrm9084:
@AlexandraKemperMS

docs/src/main/asciidoc/feature-management.adoc
```diff
@@ -0,0 +1,276 @@
+== Feature Management
+
+Feature Management provide a way for Spring Boot applications dynamically access content. Feature Management has a verity of functions, such as Feature Flags that can enable or disable content, Feature Flags for targeting when content is shown, and Feature Gates for dynamically enabling endpoints.
```
AlexandraKemperMS, line 3: The second item should be feature filters - I would add that those are based on targeted audiences, and include that we also have timewindow filters for exposing a feature for a limited amount of time.

docs/src/main/asciidoc/feature-management.adoc
```diff
@@ -0,0 +1,276 @@
...
+
+By default, loading of Feature Flags from Azure App Configuration is disabled. And is enabled through configuration:
```
AlexandraKemperMS, line 5: This makes it sound like that our customers normally do not use feature flags/this is an "out of the way" feature. Could we phrase this differently? 
Something like "To use feature flags, enable them through configuration:"

docs/src/main/asciidoc/feature-management.adoc
AlexandraKemperMS: Is this the paragraph that introduces the feature management concept to customers? Then I think we should also include the information of when to use them and explain them in more depth, you can borrow this form the [concept doc on feature management](https://docs.microsoft.com/en-us/azure/azure-app-configuration/concept-feature-management)

docs/src/main/asciidoc/feature-management.adoc
```diff
@@ -0,0 +1,276 @@
...
+
+==== Feature Flags
+
+Feature flags are composed of two parts, a name and a list of feature-filters that are used to turn the feature on. Feature Flags can either have a boolean state of on/off, or they can have a list of Feature Filters. Feature Flags evaluate Feature Filters until one returns true, if no Feature Filter returns true, then the Feature Flag returns false.
+
+==== Feature Filters
+
+Feature filters define a scenario for when a feature should be enabled. Feature Filters are evaluated synchronously.
+
+As an example, a Microsoft Edge browser feature filter could be designed. This feature filter would activate any features it is attached to as long as an HTTP request is coming from Microsoft Edge.
```
AlexandraKemperMS, line 24: Here I would mention that this would involve building a filter of type "Custom". It would be better to mention some examples of the features it is attached to - such as a feature filter that activates a seasonal marketing campaign for specifically edge customers for a set time window, which uses the timewindow filter.

docs/src/main/asciidoc/feature-management.adoc
```diff
@@ -0,0 +1,276 @@
...
+
+=== Feature Management Basics
+
+==== Feature Flags
+
+Feature flags are composed of two parts, a name and a list of feature-filters that are used to turn the feature on. Feature Flags can either have a boolean state of on/off, or they can have a list of Feature Filters. Feature Flags evaluate Feature Filters until one returns true, if no Feature Filter returns true, then the Feature Flag returns false.
+
+==== Feature Filters
+
+Feature filters define a scenario for when a feature should be enabled. Feature Filters are evaluated synchronously.
```
AlexandraKemperMS, line 22: Would mention here that we support 3 types of feature filters, the details of them can be in their own sections.

docs/src/main/asciidoc/feature-management.adoc
mrm9084: Feature Filters aren't always based on targeted audiences, I want to keep this intro generic.

docs/src/main/asciidoc/feature-management.adoc
mrm9084: @AlexandraKemperMS we do first feature return true, so your example here is actually impossible...

docs/src/main/asciidoc/feature-management.adoc
mrm9084: Updated, but it is 4 not 3.

docs/src/main/asciidoc/feature-management.adoc
```diff
@@ -0,0 +1,276 @@
...
+Feature Management provide a way for Spring Boot applications dynamically access content. Feature Management has a verity of functions, such as Feature Flags that can enable or disable content, Feature Flags for targeting when content is shown, and Feature Gates for dynamically enabling endpoints.
+
+By default, loading of Feature Flags from Azure App Configuration is disabled. And is enabled through configuration:
+
+[source,properties,indent=0]
+----
+spring.cloud.azure.appconfiguration.stores[0].feature-flags.enabled= true
+----
+
+With Feature Flags enabled they will be loaded into the Spring Configuration system with the prefix `feature-management`. Feature Flags can also be registered in the local configuration file, see <<Feature Flag Declaration>>. The easiest was to use Feature Management is by using the `azure-spring-cloud-feature-management` and `azure-spring-cloud-feature-management-web` libraries. The difference between the two libraries is that `azure-spring-cloud-feature-management-web` takes a dependency on the `spring-web` and `spring-webmvc` libraries to add additional features, such as Feature Gates.
```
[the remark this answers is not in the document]
mrm9084: Added link.

docs/src/main/asciidoc/feature-management.adoc
AlexandraKemperMS: Discussed this offline

docs/src/main/asciidoc/feature-management.adoc
AlexandraKemperMS: that is fine, then maybe call them "customized feature flags"? It is strange to read the same word defined twice in two different ways.

docs/src/main/asciidoc/feature-management.adoc
```diff
@@ -0,0 +1,279 @@
...
+}
+----
+
+=== Built-In Feature Filters
+
+There are a few feature filters that come with the `azure-spring-cloud-feature-management` package. These feature filters are not added automatically, but can be setup in an `@Configuration` for use.
+
+==== AlwaysOnFilter
+
+This filter always returns true.
```
AlexandraKemperMS, line 150: Can you mention that you have an example of how to use this in the example above? in case some didn't read the previous sections but wants an example

docs/src/main/asciidoc/feature-management.adoc
```diff
@@ -0,0 +1,279 @@
...
+
+==== PercentageFilter
+
+This filter provides the capability to enable a feature based on a set percentage.
```
AlexandraKemperMS, line 154: One question we get a lot is who this applies to - so I would add a few words that it enables the feature for everyone that has access to the filter, and that it is not sticky aka it randomly applies the percentage to a request, it does not remember who you are as a user and is not guaranteed to give you the same on or off experience consistently

docs/src/main/asciidoc/feature-management.adoc
```diff
@@ -0,0 +1,279 @@
...
+      -
+       name: TimeWindowFilter
+        parameters:
+          time-window-filter-setting-start: "Wed, 01 May 2019 13:59:59 GMT",
+          time-window-filter-setting-end: "Mon, 01 July 2019 00:00:00 GMT"
+----
+
+==== TargetingFilter
+
+This filter provides the capability to enable a feature for a target audience. An in-depth explanation of targeting is explained in the targeting section below. The filter parameters include an audience object that describes users, groups, and a default percentage of the user base that should have access to the feature. Each group object that is listed in the target audience must also specify what percentage of the group's members should have access. If a user is specified in the users section directly, or if the user is in the included percentage of any of the group rollouts, or if the user falls into the default rollout percentage, then that user will have the feature enabled.
```
AlexandraKemperMS, line 185: In the part where you mention the targeting section below, could you link to it to shortcut the experience? 

Also for the third sentence, maybe this makes it clearer:
"For each group object that is listed in the target audience, a percentage is required which defines the percentage of that group's members which has access to the feature."

docs/src/main/asciidoc/feature-management.adoc
```diff
@@ -0,0 +1,279 @@
...
+              rolloutPercentage: 100
+            -
+              name: Ring1
+              rolloutPercentage: 100
+          defaultRolloutPercentage: 50
+----
+
+=== Custom Feature Filters
+
+Creating a feature filter provides a way to enable features based on criteria that you define. To implement a feature filter, the `FeatureFilter` interface must be implemented. `FeatureFilter` has a single method `evaluate`. When a feature specifies that it can be enabled with a feature filter, the `evaluate` method is called. If `evaluate` returns `true` it means the feature should be enabled. If `false` it will continue evaluating the Feature's filters until one returns true. If all return `false` then the feature is off.
```
AlexandraKemperMS, line 210: This does not include the word custom? It sounds like you are speaking about generic feature flags - "Creating a custom feature filter"
To create a custom feature filter, the `FeatureFilter` interface must be implemented.

docs/src/main/asciidoc/feature-management.adoc
```diff
@@ -0,0 +1,279 @@
...
+
+The following steps demonstrate an example of a progressive rollout for a new 'Beta' feature:
+
+1. Individual users Jeff and Alicia are granted access to the Beta
+2. Another user, Mark, asks to opt-in and is included.
+3. Twenty percent of a group known as "Ring1" users are included in the Beta.
+4. The number of "Ring1" users included in the beta is bumped up to 100 percent.
+5. Five percent of the user base is included in the beta.
+6. The rollout percentage is bumped up to 100 percent and the feature is completely rolled out.
+7. This strategy for rolling out a feature is built into the library through the included TargetingFilter feature filter.
```
AlexandraKemperMS, line 244: Really really like this list example - only , you do not need to mention the final description line as step 7, should just be on a new line

docs/src/main/asciidoc/feature-management.adoc
```diff
@@ -0,0 +1,279 @@
...
+        name: AlwaysOnFilter
+----
+
+`feature-t` is set to false, Feature Flags defined this way just return their value. `feature-u` uses Feature Filters which are defined under `enabled-for` `feature-u` has one Feature filter, `Random`, which does not require any configuration, so it only has the name property. `feature-v` it specifies a feature filter named `TimeWindow`. Feature Filters can be passed parameters to configure the filter. In this case, the start and end times for the feature to be active are configured.
+
+The `AlwaysOnFilter` is a Filter that always evaluates as `true`. The `evaluate` field is used to stop the evaluation of the feature filters, and results in the feature flag always return `false`.
+
+=== Evaluating Feature Flags
+
+`azure-spring-cloud-feature-management` provides `FeatureManager` to check if a Feature Flag is enabled. It provides an asynchronous way to check the state of the flag.
```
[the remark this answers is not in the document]
mrm9084: added that is a library. `FeatureManager` is the name of the class.

AlexandraKemperMS (approved):

PR #886, opened 2022-03-25
Categorize configuration properties by service in the appendix

saragluna:
As title.


docs/pom.xml
```diff
@@ -104,6 +104,171 @@
...
+                </goals>
+                <configuration>
+                  <arguments>
+                    <argument>${project.basedir}/src/main/asciidoc/_configprops_kv_certificate.adoc</argument>
+                    <argument>spring.cloud.azure.keyvault.certificate.*</argument>
+                  </arguments>
+                </configuration>
+              </execution>
+              <execution>
+                <id>generate-servicebus-configprops</id>
```
yiliuTo, line 220: need we add configuration properties for jms as well?

saragluna: good point

saragluna: added

yiliuTo (approved):
LGTM

PR #892, opened 2022-03-29
upgrade spring cloud azure versions

saragluna:
Upgrade version to 4.0.0


pom.xml
```diff
@@ -42,7 +42,7 @@
          "version.spring.cloud.azure" is used to update "_configprops.adoc".
```
backwind1233, line 42: do we need to update this comments here?

saragluna: deleted

backwind1233 (approved):
LGTM

PR #900, opened 2022-03-30
Sync content from MS docs

chenrujun:
As title.

Preview page: https://chenrujun.github.io/spring-cloud-azure/4.0.0/4.0.0/reference/html/index.html


chenrujun:
Here is current sync result:
image
image

PR #918, opened 2022-04-05
App Config Docs

mrm9084:
Adding the last few sections.
Fixing a few mistakes found.

### DIFF
--- a/docs/src/main/asciidoc/client-customization.adoc
+++ b/docs/src/main/asciidoc/client-customization.adoc
@@ -1,6 +1,6 @@
 == Client Customization
 
-The library uses the https://github.com/Azure/azure-sdk-for-java[Azure SDK for Java] for connecting to Azure App Configuration and Azure Key Vault. Two interfaces, `ConfiguraitonClientBuilderSetup` and `SecretClientBuilderSetup` are provided to modify the clients. Each interface has a `setup` method that takes in their respective builder along with the `String` value of the URI that the client will connect to. 
+The App Configuration library uses the https://github.com/Azure/azure-sdk-for-java[Azure SDK for Java] for connecting to Azure App Configuration and Azure Key Vault. Two interfaces, `ConfigurationClientBuilderSetup` and `SecretClientBuilderSetup` are provided to modify the clients. Each interface has a `setup` method that takes in their respective builder along with the `String` value of the URI that the client will connect to. 
 
 [source,java,indent=0]
 ----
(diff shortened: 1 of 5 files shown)


docs/src/main/asciidoc/client-customization.adoc
```diff
@@ -0,0 +1,47 @@
+== Client Customization
+
+The library uses the https://github.com/Azure/azure-sdk-for-java[Azure SDK for Java] for connecting to Azure App Configuration and Azure Key Vault. Two interfaces, `ConfiguraitonClientBuilderSetup` and `SecretClientBuilderSetup` are provided to modify the clients. Each interface has a `setup` method that takes in their respective builder along with the `String` value of the URI that the client will connect to.
```
AlexandraKemperMS, line 3: Which library? Specify if this for App Config, otherwise this can get confusing because you have mentioned multiple libraries previously in the article

I think you mean ConfigurationClientBuilderSetup

docs/src/main/asciidoc/client-customization.adoc
```diff
@@ -0,0 +1,47 @@
...
+public interface ConfigurationClientBuilderSetup {
+    public void setup(ConfigurationClientBuilder builder, String endpoint);
+}
+
+public interface SecretClientBuilderSetup {
+    public void setup(SecretClientBuilder builder, String uri);
+}
+----
+
+This enables modifying the http client and its configurations. This can be used to replace the default `HttpClient`` with one using a proxy for all traffic to App Configuration and Key Vault.
```
AlexandraKemperMS, line 16: I would rephrase this to "These interfaces allow for customization of the http client and its configurations, for example, by replacing the default `HttpClient` with another that uses a proxy for all traffic directed to App Configuration and Key Vault. This example can be seen below"

docs/src/main/asciidoc/client-customization.adoc
```diff
@@ -0,0 +1,47 @@
...
+                new InetSocketAddress(hostname, port));
+        return new NettyAsyncHttpClientBuilder()
+                .proxy(proxyOptions)
+                .build();
+    }
+
+}
+----
+
+NOTE: The `ConfigurationClientBuilder` and `SecretClientBuilder` are already setup for use when passed into `setup`. Any changes to the clients, such as credentials and retry policy will override those already in place.
```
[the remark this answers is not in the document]
AlexandraKemperMS: I would add this before the sample use case text, since this is something a reader would want to know when looking at lines 5-12

docs/src/main/asciidoc/refresh.adoc
```diff
@@ -30,7 +30,7 @@ The App Configuration Spring libraries support the ability to periodically check
 
 [,properties,indent=0]
 ----
-spring.cloud.azure.appconfiguration.stores[0].monitoring.refresh-interval= 1d
+spring.cloud.azure.appconfiguration.stores[0].monitoring.refresh-interval= 30s
```
AlexandraKemperMS, line 33: Can we set this to a longer period, like 1 minute or 5 minutes.  We had an issue in our azure docs that customer would set a refresh interval of 30s and we saw more customers running though their request quota because they never changed this later on

mrm9084: Sure, I only changed it to that, as it was the default for this configuration. I was getting questions about it.

AlexandraKemperMS: Ah ok, if customers were asking questions about it and it is the default, then we can also keep it, your choice. Spring customers are not always expecting the same things as the ones in the rest of our docs.

AlexandraKemperMS (approved):

PR #920, opened 2022-04-13
Fix appendix title error.

hui1110:
Modify the `proeprties` in the title of the [current page](https://microsoft.github.io/spring-cloud-azure/current/reference/html/appendix.html) to `properties`.


saragluna (approved):
LGTM.

PR #938, opened 2022-05-09
Update README.md.

chenrujun:
This repo is high possibility appeared in the result when searching "Azure Spring Cloud" by searching engine. 

We should put more information about "Spring on Azure" project instead of just let customer know that this repo is used to manage the reference doc.

So. I add these items in the README.md:
1. Brief introduction of "Spring on Azure" introduction.
2. Links to reference doc / samples / source code.

And I already updated the "About" fragment in current repo:
image


saragluna (approved):

saragluna:
Should we also provide link to the ms docs?

PR #948, opened 2022-05-13
Fix pipeline broken link checker error

hui1110:
GitHub action file: https://github.com/microsoft/spring-cloud-azure/blob/main/.github/workflows/brokenLinkCheck.yml

In step `Link version replacement`, when there is a broken link, it will be output to link-checker/out.md. When there is no broken link, it is not processed in the current pipeline.


chenrujun (approved):
LGTM.

PR #949, opened 2022-05-13
Update markdown link checker

hui1110:
GitHub action file: https://github.com/microsoft/spring-cloud-azure/blob/main/.github/workflows/brokenLinkCheck.yml

In step `Check link in all markdown files`, when the link ends with more than one `)`, the action check result is incorrect. Replace actions that can be checked normally. For example:

Markdown file: https://github.com/microsoft/spring-cloud-azure/blame/main/SECURITY.md#L7
check result: https://github.com/microsoft/spring-cloud-azure/runs/6422354007?check_suite_focus=true#step:8:34


.github/workflows/brokenLinkCheck.yml
```diff
@@ -59,11 +59,13 @@ jobs:
              sudo mv out.md ../
           fi
 
+      - name: Remove markdown files that don't check for links
```
chenrujun, line 62: Remove files generated at build-time.

chenrujun (changes requested):

chenrujun (approved):
LGTM.

Issue #937, opened 2022-05-06
Storage fileshare donot support TokenCredential


jialigit:
## Context
As per our [reference doc of 4.0.0](https://microsoft.github.io/spring-cloud-azure/4.0.0/reference/html/appendix.html#_azure_storage_file_share_proeprties), in the appendix section for storage fileshare, credential configuration is not needed.

## Goal
We should remove the credential related option for fileshare section.

PR #976, opened 2022-07-05
not include credential options for fileshare namespace

jialigit:
As title.

solves issue:
https://github.com/microsoft/spring-cloud-azure/issues/937

### DIFF
--- a/docs/src/main/asciidoc/_configuration-properties-all.adoc
+++ b/docs/src/main/asciidoc/_configuration-properties-all.adoc
@@ -1140,13 +1140,6 @@
 |spring.cloud.azure.storage.fileshare.client.response-timeout |  | Amount of time used when waiting for a server to reply.
 |spring.cloud.azure.storage.fileshare.client.write-timeout |  | Amount of time each request being sent over the wire.
 |spring.cloud.azure.storage.fileshare.connection-string |  | Connection string to connect to the service.
-|spring.cloud.azure.storage.fileshare.credential.client-certificate-password |  | Password of the certificate file.
-|spring.cloud.azure.storage.fileshare.credential.client-certificate-path |  | Path of a PEM certificate file to use when performing service principal authentication with Azure.
-|spring.cloud.azure.storage.fileshare.credential.client-id |  | Client ID to use when performing service principal authentication with Azure.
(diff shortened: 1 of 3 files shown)


hui1110:
Hi @jialigit, will the property `spring.cloud.azure.storage.fileshare.credential.*` be in the file `_configuration-properties-all.adoc`? 
This PR should contain the file `_configuration-properties-azure-storage-file-share. adoc` that was changed after executing the command `mvn clean install -Pdoc`. If the property `spring.cloud.azure.storage.fileshare.credential.*` will not be in the file `_configuration-properties-all.adoc` then it should also be included after the update.

jialigit:
> 

OK, Let's use only one PR to update them all at once.

hui1110:
Hi @jialigit, the [pr](https://github.com/MicrosoftDocs/azure-dev-docs-pr/pull/3168) to update `spring-cloud-azure-appendix` in MS Docs has been merged, please create PR update [spring-cloud-azure-appendix](https://docs.microsoft.com/en-us/azure/developer/java/spring-framework/spring-cloud-azure-appendix).

chenrujun (approved):
1. LGTM about "Removing credential options in appendix".
2. As discussed, MS docs is the single source of truth, this should be merged after MS docs updated. But if this PR does not merge first, the MS docs will not be updated. So it's OK to merge this PR first.
3. If these properties don't take effect, we should avoid these properties from Java code. Not just removing these properties in the reference doc. It's like a bug of source code, but we hide this bug in reference doc.

jialigit:
> 

Yes. This properties(AzureStorageFileShareProperties) inherit from the parent properties, but some of them are not available for this service. As for now, we can't modify the structure of the class which will bring breaking changes to the API, so here we just only need to suppress the properties file generation rule.

jialigit:
This is the PR update the options in MS doc: https://github.com/MicrosoftDocs/azure-dev-docs-pr/pull/3279

We will merge the update into MS doc first mannualy, and then merge this PR into spring-cloud-azure repo. But the azure-dev-docs-pr repo may not merge it very quickly, so this PR is likely be blocking here for a few days.

jialigit:
> 

As PR : [#3279](https://github.com/MicrosoftDocs/azure-dev-docs-pr/pull/3279) is merged, we merge this to sync the changes.

PR #1002, opened 2022-09-26
Update version to 4.5.0-beta.1

Netyyyy:
As title.


chenrujun:
We should merge it after this new tag created: spring-cloud-azure_4.4.0

saragluna:
@Netyyyy since @hui1110 will take care of this, you could close this PR now.

PR #1003, opened 2022-09-26
Update current version to 4.4.0

Netyyyy:
As title.


current
```diff
@@ -1 +1 @@
-4.3.0
\ No newline at end of file
+4.5.0-beta.1
```
chenrujun, line 1: Current version should be latest released version. It should be `4.4.0`

chenrujun (changes requested):

PR #1009, opened 2022-09-27
Add logic to generate beta version documentation

hui1110:
As title.

### DIFF
--- a/docs/src/main/asciidoc/ghpages.sh
+++ b/docs/src/main/asciidoc/ghpages.sh
@@ -154,17 +154,17 @@ function copy_docs_for_current_version() {
             fi
         done
         COMMIT_CHANGES="yes"
-    elif [[ "${BETA_RELEASE_BRANCHES_VALUE}" == "feature/spring-cloud-azure-passwordless-connection" ]] ; then
-      echo -e "Current branch is '${BETA_RELEASE_BRANCHES_VALUE}' - will copy the current docs only to the root folder"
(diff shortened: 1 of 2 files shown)


docs/src/main/asciidoc/ghpages.sh
```diff
@@ -145,6 +154,17 @@ function copy_docs_for_current_version() {
             fi
         done
         COMMIT_CHANGES="yes"
+    elif [[ "${BETA_RELEASE_BRANCHES_VALUE}" == "feature/spring-cloud-azure-passwordless-connection" ]] ; then
```
saragluna, line 157: here we should check wether current branch is one of the BETA_RELEASE_BRANCHES_VALUE

docs/src/main/asciidoc/ghpages.sh
```diff
@@ -145,6 +154,17 @@ function copy_docs_for_current_version() {
...
+      echo -e "Current branch is '${BETA_RELEASE_BRANCHES_VALUE}' - will copy the current docs only to the root folder"
```
saragluna, line 158: Current branch is `CURRENT_BRANCH`

saragluna: The log info should be like, 

```
Branch [${CURRENT_BRANCH}] is a whitelisted beta release branch! Will copy the current docs to the root folder

```

saragluna (approved):

yiliuTo (approved):
LGTM

PR #1011, opened 2022-09-27
Add logic to generate beta version documentation in feature branch

hui1110:
As title.

### DIFF
--- a/docs/pom.xml
+++ b/docs/pom.xml
@@ -16,10 +16,9 @@
   <description>Spring Cloud Azure Build Docs</description>
 
   <properties>
-    <docs.main>spring-cloud-azure</docs.main>
     <main.basedir>${basedir}/..</main.basedir>
-    <docs.whitelisted.branches>${project.version}</docs.whitelisted.branches>
-    <docs.beta.release.branches>feature/spring-cloud-azure-passwordless-connection</docs.beta.release.branches>
+    <release.to.gh.pages.enabled>true</release.to.gh.pages.enabled>
+    <release.to.gh.pages.target.floder>/</release.to.gh.pages.target.floder>
     <configprops.inclusionPattern>spring.jms.servicebus.*|spring.cloud.azure.*</configprops.inclusionPattern>
     <upload-docs-zip.phase>none</upload-docs-zip.phase>
     <generated-docs-multipage-output.dir>${project.build.directory}/generated-docs/${project.version}/reference/html</generated-docs-multipage-output.dir>
--- a/docs/pom.xml
+++ b/docs/pom.xml
@@ -19,6 +19,7 @@
     <docs.main>spring-cloud-azure</docs.main>
     <main.basedir>${basedir}/..</main.basedir>
     <docs.whitelisted.branches>${project.version}</docs.whitelisted.branches>
+    <docs.beta.release.branches>feature/spring-cloud-azure-passwordless-connection</docs.beta.release.branches>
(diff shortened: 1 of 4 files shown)


saragluna (approved):

docs/pom.xml
```diff
@@ -19,6 +19,7 @@
     <docs.main>spring-cloud-azure</docs.main>
     <main.basedir>${basedir}/..</main.basedir>
     <docs.whitelisted.branches>${project.version}</docs.whitelisted.branches>
+    <docs.beta.release.branches>feature/spring-cloud-azure-passwordless-connection</docs.beta.release.branches>
```
chenrujun, line 22: Why create a new property? 
1. `docs.whitelisted.branches` can be modified to the feature name.
2. `docs.whitelisted.branches` support a list. We can do like this: `<docs.whitelisted.branches>${project.version}, feature/spring-cloud-azure-passwordless-connection</docs.whitelisted.branches>`

chenrujun (changes requested):

docs/src/main/asciidoc/ghpages.sh
```diff
@@ -145,6 +154,17 @@ function copy_docs_for_current_version() {
...
         done
         COMMIT_CHANGES="yes"
+    elif [[ ",${BETA_RELEASE_BRANCHES_VALUE}," = *",${CURRENT_BRANCH},"* ]] ; then
+        echo -e "Branch [${CURRENT_BRANCH}] is a whitelisted beta release branch! Will copy the current docs to the root folder"
+        for f in docs/target/generated-docs/*; do
+            file=${f#docs/target/generated-docs/*}
+            if ! git ls-files -i -o --exclude-standard --directory | grep -q ^$file$; then
+                # Not ignored...
+                cp -rf $f ${ROOT_FOLDER}/
+                git add -A ${ROOT_FOLDER}/$file
```
chenrujun, line 164: nit: wrap this logic into a function.

docs/src/main/asciidoc/ghpages.sh
```diff
@@ -145,6 +154,17 @@ function copy_docs_for_current_version() {
...
+            fi
+        done
+        COMMIT_CHANGES="yes"
     else
         echo -e "Current branch is [${CURRENT_BRANCH}]"
```
chenrujun, line 169: If we always use `${version}` instead of `${branch}/${version}`, we can delete all logic about `${branch}/${version}.`

docs/src/main/asciidoc/ghpages.sh
```diff
@@ -134,44 +135,17 @@ function add_docs_from_target() {
 
 # Copies the docs by using the retrieved properties from Maven build
 function copy_docs_for_current_version() {
-    if [[ "${CURRENT_BRANCH}" == "main" ]] ; then
-        echo -e "Current branch is main - will copy the current docs only to the root folder"
+    if [[ "${RELEASE_TO_GH_PAGES_ENABLED_VALUE}" == "true" ]] ; then
+        echo -e "Need to release - will copy the current docs only to the root folder"
```
chenrujun, line 139: `will copy the current docs only to the root folder` -> `Will copy the all files from "docs/target/generated-docs/ to ${ROOT_FOLDER}${RELEASE_TO_GH_PAGES_TARGET_FLODER_VALUE} "`

chenrujun (changes requested):

chenrujun (approved):

PR #1020, opened 2022-10-09
Remove unwanted trigger branches from actions that generate documentation

hui1110:
As title.

### DIFF
--- a/.github/workflows/update-docs.yaml
+++ b/.github/workflows/update-docs.yaml
@@ -4,8 +4,6 @@ on:
   push:
     branches:
     - main
-    - docs
-    - 4.*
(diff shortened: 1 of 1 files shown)


.github/workflows/update-docs.yaml
```diff
@@ -6,7 +6,6 @@ on:
     - main
     - docs
     - 4.*
-    - feature/spring-cloud-azure-passwordless-connection
```
chenrujun, line 9: Does it mean we can only keep the branch itself for each branch?
For example: In `main` branch, `on.push.branches = main`.

chenrujun (approved):

PR #1067, opened 2024-12-19
Update generated description

Netyyyy:
Replace single/double quotes with backticks, ignoring possessive and nested cases.


Netyyyy:
@saragluna please help take a review

spring-cloud-azure-markdown-appdendix-generator/src/main/java/com/azure/spring/cloud/internal/Main.java
saragluna, line 1: Do you think we should replace double quote as well?

saragluna (approved):

PR #1068, opened 2024-12-25
Update url and some formats

Netyyyy:
change `xxx: https://learn.microsoft.com/xxx` to `[xxx](xxx)`
add `` for GrantedAuthority's and access_token's
fix not to change n't


saragluna (approved):

spring-cloud-azure-markdown-appdendix-generator/src/main/java/com/azure/spring/cloud/internal/Main.java
```diff
@@ -154,6 +154,16 @@ protected List<Resource> getSpringConfigurationMetadataJsonFilesInClasspath() th
 			}).collect(Collectors.toList());
 		}
 
+		private String formatDescription(String description) {
```
saragluna, line 157: Is this hard coded here? If we have another word, will this still work?

